PR #3, opened 2018-03-28
Add BuildFrom

julienrf:
The goal of this PR is to make it possible for library authors to write code that uses the new `BuildFrom` but that still works with old Scala versions.

This PR introduces the `BuildFrom` type and derives instances from existing `CanBuildFrom` instances.

### DIFF
--- a/src/test/scala/collection/CollectionTest.scala
+++ b/src/test/scala/collection/CollectionTest.scala
@@ -1,3 +1,5 @@
+package collection
+
 import java.util
 
 import org.junit.Test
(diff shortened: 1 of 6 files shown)


julienrf:
The CI failed because it requires #4 to be merged first.

src/test/scala/collection/CollectionTest.scala
```diff
@@ -1,3 +1,5 @@
+package collection
```
szeiger: What's the point of putting the tests into this package?

julienrf: Just to not use the empty package.

PR #7, opened 2018-03-28
Add an `iterator()` alias to `iterator`

julienrf:
We already have a [scalafix rule](https://github.com/scala/collection-strawman/blob/master/scalafix/2.13/rules/src/main/scala/fix/Collectionstrawman_v0.scala#L36-L37) that translates `.iterator` to `.iterator()`. This rule can be applied to all the code bases to make them compatible with 2.13. However, once the rule has been applied, the code won’t be compatible with 2.12 (or previous Scala version) anymore…

This PR introduces an `iterator()` extension method for Scala 2.12 so that after the application of the scalafix rule, the code can still compile with 2.12.

Also, related discussion: https://github.com/scala/collection-strawman/issues/520


julienrf:
I’m on the fence with the changes on parentheses. On one hand, there are *some* advantages of having a more disciplined used of parentheses (although, as noted by @Ichoran in the linked discussion, it is hard to find a really robust principle on when to use parentheses). But on the other hand that a source of incompatibilities between  2.13 and prior Scala versions… The trick I’m using here works for `iterator()` but it doesn’t always work for `newBuilder()`, because sometimes this `newBuilder()` method takes implicit parameters (typically, an `Ordering` or a `ClassTag`), and in that case the implicit conversion to the extension method is not triggered.

src/main/scala-2.12/collection/compat/package.scala
```diff
@@ -30,15 +30,31 @@ package object compat {
     simpleCBF(fact.newBuilder)
 
   implicit class IterableFactoryExtensionMethods[CC[X] <: GenTraversable[X]](private val fact: GenericCompanion[CC]) {
-    def from[A](source: TraversableOnce[A]): CC[A] = fact.apply(source.toSeq: _*)
+    def from[A](source: TraversableOnce[A]): CC[A] = (fact.newBuilder[A] ++= source).result()
```
szeiger: This won't work for lazy collections. You don't want `Stream.from(otherStream)` to force anything.

PR #8, opened 2018-03-28
Add scala.collection.Factory

julienrf:


src/main/scala-2.12/collection/Factory.scala
```diff
@@ -0,0 +1,36 @@
...
+  /** Get a Builder for the collection. For non-strict collection types this will use an intermediate buffer.
+    * Building collections with `fromSpecific` is preferred because it can be lazy for lazy collections. */
+  def newBuilder(): mutable.Builder[A, C]
+}
+
+object Factory {
+
+  implicit def fromCanBuildFrom[A, C](implicit cbf: CanBuildFrom[Nothing, A, C]): Factory[A, C] =
+    new Factory[A, C] {
+      def fromSpecific(it: TraversableOnce[A]): C = (cbf() ++= it).result()
```
szeiger: This needs special treatment of `Stream` and `LazyList` to preserve laziness.

julienrf: Actually this works because `Stream` uses a `LazyBuilder`. I’ve added a test for that.

PR #21, opened 2018-05-05
newer 2.13 snapshot, 2.12.6, fix language warnings

SethTisue:
the goal here is to get the project green again in the 2.13 community build.

(the 2.12.6 and language warnings parts are just gravy.)

(despite the branch name, this does not enable fatal warnings, since of course on 2.13 we have deprecation warnings)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,8 @@
 language: scala
 scala:
   - 2.11.12
-  - 2.12.5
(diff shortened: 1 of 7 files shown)


SethTisue:
> the goal here is to get the project green again in the 2.13 community build

verified in local run that these changes do the trick.

PR #22, opened 2018-05-08
Add knownSize and iterableFactory to old collections

szeiger:


julienrf:
We should also have corresponding migration rules rewriting `.companion` to `.iterableFactory`.

MasseGuillaume:
@julienrf let's close this PR in favor of https://github.com/scala/scala-collection-compat/pull/143

.companion => .iterableFactory is already in master: https://github.com/scala/scala-collection-compat/pull/116

Issue #16, opened 2018-04-20
Setup release infrastructure


julienrf:

lrytz:
I can take care of that, at least for the compat library part. We have some useful scripts in other repos to do tag driven releases on travis.

julienrf:
Thanks Lukas! In case you were wondering there is nothing to do to “release” the Scalafix migration rule because Scalafix can download it directly from GitHub. This issue is just about releasing the `scala-collection-compat` artifact.

lrytz:
Thanks for raising the issue! Agree on all points.

> we can't publish collections-compat until after Scala.js had been published

We could publish the JVM version earlier in case Scala.js for M4 takes a while to be ready.

PR #23, opened 2018-05-09
Infrastructure for tag-based publishing

lrytz:
Fixes https://github.com/scala/scala-collection-compat/issues/16


lrytz:
@SethTisue I moved away from using the rather brittle `scalaVersionsByJvm` feature of the sbt module plugin. Instead, the travis build matrix sets the combinations of scala and jvm versions to test, and for those combinations that should be used in a release, it needs to make sure the `RELEASE_COMBO` var is `true`. Let me know what you think. Maybe it can be further simplified. I'd like to do something like this for the other modules as well.

lrytz:
Tested releasing by creating a dummy tag (https://github.com/scala/scala-collection-compat/releases/tag/v0.0.1-TEST). It created 4 staging repos (it's still misterious to me why / how releases are split into multiple staging repos...)
  - https://oss.sonatype.org/content/repositories/orgscala-lang-1544/org/scala-lang/modules/
  - https://oss.sonatype.org/content/repositories/orgscala-lang-1545/org/scala-lang/modules/
  - https://oss.sonatype.org/content/repositories/orgscala-lang-1546/org/scala-lang/modules/
  - https://oss.sonatype.org/content/repositories/orgscala-lang-1547/org/scala-lang/modules/

So I think this is good to go.

project/plugins.sbt
```diff
@@ -3,3 +3,4 @@ val scalaJSVersion =
 
 addSbtPlugin("org.scala-js" % "sbt-scalajs" % scalaJSVersion)
 addSbtPlugin("org.portable-scala" % "sbt-scalajs-crossproject" % "0.4.0")
+addSbtPlugin("org.scala-lang.modules" % "sbt-scala-module" % "1.0.12")
```
SethTisue: should be 1.0.14?

lrytz: Sure!

PR #24, opened 2018-05-15
ImmutableArray -> ArraySeq, iterator() -> iterator

lrytz:


lrytz:
We should also get a rewrite rule for `ArrayBuilder.make()` -> `ArrayBuilder.make`.

Issue #30, opened 2018-05-16
scala-collection-compat_2.12-0.1.0.jar does not contains scala.collection.BuildFrom due to osgi setting


xuwei-k:
https://github.com/scala/scala-collection-compat/blob/5d1f7922ef25211938290b83fb9817e8a7e677c9/build.sbt#L27-L28

lrytz:
Oh, it threw out everything that's not in the compat package... Too bad I didn't notice when looking at the jar.

lrytz:
Thinking how to go forward on this issue, I think it would be best to get rid of the split packages. Instead, the `scala-collection-compat` library should only expose types in the `scala.collection.compat` package.

This means we'd put the `Factory` and `BuildFrom` and `ArraySeq` (maybe `collection.compat.immutable.ArraySeq`?) classes into the compat package. In the 2.13 version, they would be type aliases.

Code that cross-compiles would use `collection.compat.Factory`. Once the library moves to 2.13 only, the code can change to `collection.Factory`.

julienrf:
What is the purpose of this osgi setting?

lrytz:
To use our jars with osgi they need to define metadata. Split packages are problematic, search on google search. I personally have no experience with osgi, few people around me do. But the metadata is improtant for those that want to use it. Our working mode is to continue emitting the metadata, rely on the community to report and fix issues.

In any case, split packages will also cause problems with the java module system.

julienrf:
Ok, I think your solution would work well.

lrytz:
Note: the idea of putting `ArraySeq` into the `collection.compat.immutable` package is not a good one. Having two `immutable` packages has too much potential to cause conflicts and confusion.

julienrf:
The only problematic case I can think of is :

import collection.immutable
import collection.compat._

immutable.Something

Because in the last line "immutable" confusingly refers to the one in the compact object.

We can document this wart. I don't expect people to often import collection.immutable (in contrast with collection.mutable).

szeiger:
I don't know how much of a problem this really is for OSGi. The library is coupled tightly to the standard library anyway so a split package might be just fine. AFAIK Java 9 modules don't support split packages though, so we may have to change it for that reason alone (i.e. if we want to support modules).

lrytz:
I'm really not sure we should reuse `immutable`. I think we cannot just assume that noone imports `collection.immutable`.

```
scala> :pa -raw
// Entering paste mode (ctrl-D to finish)

[...]

PR #31, opened 2018-05-17
Move everything into the scala.colleciton.compat package

lrytz:
Fixes https://github.com/scala/scala-collection-compat/issues/30

~~Using `scala.collection.compat.imm` (not `.immutable`) to avoid
name clashes with `scala.collection.immutable`.~~

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -22,9 +22,6 @@ lazy val `scala-collection-compat` = crossProject(JSPlatform, JVMPlatform)
     }
   )
   .jvmSettings(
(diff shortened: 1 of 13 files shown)


lrytz:
Just found out that scala-parallel-collections uses `parallel.mutable/immutable`. It seems to be OK in practice, so I'll do the same here.

I also found out that parallel collections exports the `scala.collection` with a `split-package` clause, we could probably do that as well. But again, in the interest of the Java 9 module system, it's better not to do it.

julienrf:
OK to use `compat.immutable`.

PR #34, opened 2018-06-11
Rewrite `retain` to `filterInPlace`

MasseGuillaume:
fix: https://github.com/scala/collection-strawman/issues/556

### DIFF
--- a/scalafix/rules/src/main/scala/fix/Scalacollectioncompat_NewCollections.scala
+++ b/scalafix/rules/src/main/scala/fix/Scalacollectioncompat_NewCollections.scala
@@ -30,6 +30,17 @@ case class Scalacollectioncompat_NewCollections(index: SemanticdbIndex)
     Symbol("_root_.scala.runtime.Tuple3Zipped.Ops.zipped.")
   )
 
+  val retain = 
+    SymbolMatcher.normalized(
+      Symbol("_root_.scala.collection.mutable.MapLike.retain.")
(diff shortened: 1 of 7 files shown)


julienrf:
There is a subtlety: in 2.12 `retain` takes a `Function2[K, V, Boolean]`, whereas in 2.13 `filterInPlace` takes a `Function1[(K, V), Boolean]`.

julienrf:
The code generated by the rewrite should compile with `-Yno-adapted-args`.

julienrf:
Then I guess it becomes a bit complicated, you will have to transform a `(x, y) => e` expression into `{ case (x, y) => e }`. Is it possible to achieve?

scalafix/rules/src/main/scala/fix/Scalacollectioncompat_NewCollections.scala
```diff
@@ -30,6 +30,28 @@ case class Scalacollectioncompat_NewCollections(index: SemanticdbIndex)
     Symbol("_root_.scala.runtime.Tuple3Zipped.Ops.zipped.")
   )
 
+  val retain = 
+    SymbolMatcher.normalized(
+      Symbol("_root_.scala.collection.mutable.MapLike.retain.")
```
julienrf: Oops, I’ve just noticed that this operation is also on `mutable.Set`. We should support that use case too…

PR #38, opened 2018-06-11
make build work on Java 9

SethTisue:
for the Scala 2.12 community build on Java 9, not otherwise essential


SethTisue:
(we have already made the same change in other module repos such as scala-parser-combinators)

PR #40, opened 2018-06-13
Build the 2.11 version on Java 6

lrytz:
And a few small cleanups to the build scripts to bring them in line with the other repos (currently scala-xml)


lrytz:
Seems to work as expected. Review by @julienrf

PR #45, opened 2018-06-18
support for scala-native

MasseGuillaume:


sjrd:
Introducing a dependency on uTest is very dangerous. First, it prevents uTest itself from ever using scala-collection-compat. But more critically, it will introduce further delay on the republishing of the ecosystem for new versions of Scala. Now virtually *everyone* is going to transitively wait on uTest.

It's cool that the migration of the tests was done to demonstrate that scala-collection-compat actually works with Scala Native. But given the downsides of that migration, I would advocate that, in the main build, the Scala Native version be only built and published, but not tested. We can keep a separate branch with the migrated tests, to be rebased every now and then to ensure Scala Native keeps working.

MasseGuillaume:
@sjrd makes sense, we should compile + link on native, right now ArraySeq won't link.

Eventually, a `ScalaNativeJUnitPlugin` would be useful.

MasseGuillaume:
related: https://github.com/scala-native/scala-native/issues/1256

PR #48, opened 2018-06-19
bump scalafix to 0.6.0-M8

MasseGuillaume:


MasseGuillaume:
@julienrf this one is good to go.

julienrf:
Great, is a 0.6.0 planned to be released soon?

Also, I think this PR is a good opportunity to clearly mention in the documentation which scalafix version is supported. Which scala versions are supported by 0.6.0-M8, by the way?

olafurpg:
Given this PR only changes the g8 build (it doesn't use any v0.6 features) I recommend staying on v0.5.10 for a while longer. The scalafix website currently shows up-to-date docs for v0.5.10.

julienrf:
I agree with @olafurpg, we should delay the migration to 0.6.0.

olafurpg:
I think it's fine to continue working on the migration rules using the latest milestone but I recommend not "publishing" it by merging until v0.6 stable is out.

PR #53, opened 2018-06-22
Rewrite `immutable.Set/Map.+`

MasseGuillaume:
The `+` operation no longer has an overload accepting multiple values

motivation:
https://github.com/scala/collection-strawman/wiki/FAQ#what-are-the-breaking-changes

2nd row, breaking change

### DIFF
--- a/scalafix/rules/src/main/scala/fix/Scalacollectioncompat_newcollections.scala
+++ b/scalafix/rules/src/main/scala/fix/Scalacollectioncompat_newcollections.scala
@@ -165,15 +165,23 @@ case class Scalacollectioncompat_newcollections(index: SemanticdbIndex)
 
   def replaceSetMapPlus2(ctx: RuleCtx): Patch = {
     def rewritePlus(ap: Term.ApplyInfix, lhs: Term, op: Term.Name, rhs1: Term, rhs2: Term): Patch = {
-      ctx.replaceTree(
-        ap,
+
+      val tokensToReplace =
+        if(ap.tokens.headOption.map(_.is[Token.LeftParen]).getOrElse(false)) {
+          // don't drop surrounding parens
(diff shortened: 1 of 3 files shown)


scalafix/rules/src/main/scala/fix/Scalacollectioncompat_newcollections.scala
```diff
@@ -156,6 +163,28 @@ case class Scalacollectioncompat_newcollections(index: SemanticdbIndex)
...
+  def replaceSetMapPlus2(ctx: RuleCtx): Patch = {
+    def rewritePlus(ap: Term.ApplyInfix, lhs: Term, op: Term.Name, rhs1: Term, rhs2: Term): Patch = {
+      ctx.replaceTree(
+        ap,
+        Term.ApplyInfix(
+          Term.ApplyInfix(lhs, op, Nil, List(rhs1)),
+          op,
+          Nil,
+          List(rhs2)
+        ).toString
```
julienrf: What does `toString` do? That looks suspect :smile:

MasseGuillaume: It's fine, it will drop any comments. It's the scope of https://github.com/scalacenter/scala-syntax

julienrf: I was wondering whether `toString` was the recommended way to produce scala source code from a tree or if there was a method with a more meaningful name.

MasseGuillaume:
@julienrf I don't see any corner case. `.syntax` takes care of everything. I added a test + fix for `(set + (2, 3)).map(x => x)` it was removing the `()` from the lhs. Nice catch!

julienrf:
Thanks!

PR #59, opened 2018-06-25
Rewrite Iterable.sameElements -> Iterable.iterator.sameElements

MasseGuillaume:


julienrf:
Oh, good catch, I wasn’t able to find where the failure was coming from. Yeah, feel free to open a PR to fix the CI!

Issue #57, opened 2018-06-25
Support all breaking changes listed in the FAQ


julienrf:
See: https://github.com/scala/collection-strawman/wiki/FAQ#what-are-the-breaking-changes

There are still some breaking changes for which we don’t have a migration rule.

MasseGuillaume:
* [ ] _

```diff
-xs.to[List]
+xs.to(List)
```

* [ ] _

```diff
-Set(1) + (2, 3)
+Set(1) + 2 + 3
```

* [ ] _

```diff
-mutable.Set(1) + 2
+mutable.Set(1).clone() += 2
```

* [ ] _

```diff
-mutable.Map(1 -> 2).updated(1, 3)
+mutable.Map(1 -> 2).clone() += 1 -> 3
```

* [ ] _

```diff
-xs + 1 - 2
+xs ++ Set(1) -- Set(2)
```

* [ ] _

```diff
-kvs.mapValues(f)
+kvs.mapValues(f).toMap
```

* [ ] _

```diff
-xs1.sameElements(xs2)
+xs1.iterator.sameElements(xs2)  
```

* [ ] _

```diff
LinearSeq(1, 2, 3)
List(1, 2, 3) 
```

* [ ] _

```diff
map.zip(iterable)
map.zip(iterable).toMap 
```

* [ ] _

```diff
-val xs: List[Int] = ys.map(f)(collection.breakOut)
+val xs = ys.view.map(f).to(List)
```

* [ ] SeqForwarder, MutableList, StreamView no longer exist     
* [ ] collection.Map no longer have -- operation

julienrf:
@MasseGuillaume Following up on the discussion in #58 and #62, I suggest that we put aside these two items for now and we move forward on our next goals. What do you think?

Issue #56, opened 2018-06-22
Modularize the rewrite rules (move out those that don’t produce semantically equivalent code)


julienrf:
So that users can have fine-grained control on which rules to apply if they want to. In particular, one possibly controversial migration is the replacement of `Stream` with `LazyList`, since they can produce different behaviors. We should isolate the “`Stream` => `LazyList`” rule (we can just name it `StreamToLazyList`).

See the related discussion here: https://github.com/scala/collection-strawman/pull/528#issuecomment-378255181

PR #60, opened 2018-06-25
Rewrite Map.mapValues -> Map.mapValues.toMap

MasseGuillaume:


MasseGuillaume:
@julienrf one step at a time. Let's implement everything in #57 then I can split them in #56

julienrf:
I think it’s simpler to first address #56 but I’m fine as long as both approaches are eventually consistent.

julienrf:
I will merge this one, can you fix the conflict?

MasseGuillaume:
:heavy_check_mark:

PR #58, opened 2018-06-25
Rewrite collection.{Map, Set}.+ and collection.Set.-

MasseGuillaume:
This is incomplete since, it's blocked by https://github.com/scalameta/scalameta/issues/1212 (Add support to query type of a term)

iset: immutable.Set[Int]
cset: collecion.Set[Int]

both have +/- implemented via SetLike

given

iset + 1
cset + 1

we know that + is from SetLike, but it's not possible to get the type of iset/cset

PR #66, opened 2018-06-27
Experimental

MasseGuillaume:
superseded: https://github.com/scala/scala-collection-compat/pull/62 and https://github.com/scala/scala-collection-compat/pull/58

### DIFF
--- a/scalafix/rules/src/main/scala/fix/Scalacollectioncompat_newcollections.scala
+++ b/scalafix/rules/src/main/scala/fix/Scalacollectioncompat_newcollections.scala
@@ -49,6 +49,8 @@ case class Scalacollectioncompat_newcollections(index: SemanticdbIndex)
     Symbol("_root_.scala.Predef.Map#")
   )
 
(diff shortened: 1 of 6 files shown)


MasseGuillaume:
@julienrf good to go.

scalafix/rules/src/main/scala/fix/Scalacollectioncompat_newcollections.scala
```diff
@@ -26,6 +26,31 @@ case class Scalacollectioncompat_newcollections(index: SemanticdbIndex)
...
+      index.denotation(tree)
+           .exists(_.names.headOption.exists(n => symbols.exists(_ == n.symbol)))
+    }
+  }
+
+  val CollectionMap: TypeMatcher = TypeMatcher(
+    Symbol("_root_.scala.collection.immutable.Map#"),
+    Symbol("_root_.scala.collection.mutable.Map#"),
+    Symbol("_root_.scala.Predef.Map#")
+  )
```
julienrf: Eventually we should share this code instead of duplicating it. That can be done in #56, though.

PR #72, opened 2018-06-28
Major Rewrite refactor: fix #19, fix #56

### LINKED ISSUE #19: Add `CrossCompat` migration rule that produces code which cross-compiles with 2.12-
The `NewCollections` migration rule produces code that only compiles with Scala 2.13.0. This is fine for applications, but libraries usually cross compile to several Scala versions.

We should add another migration rule (e.g. `CrossCompat`), which would produce code that happily cross-compiles with all the Scala versions supported by the `scala-collection-compat` module.

Compared to the `NewCollections` rule, the `CrossCompat` rule should:
- apply a subset of the rewrites that produces code compiling with 2.12-, or code supported by the compat library,
- add an import to `scala.collection.compat._` when necessary.

### LINKED ISSUE #56: Modularize the rewrite rules (move out those that don’t produce semantically equivalent code)

### LINKED ISSUE #73: Publish rewrite rules on maven central
The current publishing mechanism with GitHub restricts us in writing rewrite rules on a single file.

MasseGuillaume:
* Move scalafix in the main build
* Add `CrossCompat` migration rule that produces code which cross-compiles with 2.12 (fix #19)
* Move LazyList and Map.mapValue rules to separate rules (fix #56) 
* We must publish to maven central now: (fix #73)


scalafix-output212/src/main/scala/fix/CanBuildFromSrc.scala
```diff
@@ -0,0 +1,36 @@
+
+
+
+package fix
```
MasseGuillaume: @julienrf :tada: cross-compatible CanBuildFrom migration.

MasseGuillaume:
@julienrf this one is good to go. Let's review it together.

CONTRIBUTING.md
```diff
@@ -29,18 +29,18 @@ Even better, instead of providing a diff, you can directly add it as a test case
 
 1. Fork this repository and create a separate branch;
 
-2. Add a file in the `scalafix/input/src/main/scala/fix/` directory with code
+2. Add a file in the `scalafix-input/src/main/scala/fix/` directory with code
```
julienrf: You should fix these references too

julienrf:
@MasseGuillaume The CI fails with this error: “sbt.ResolveException: unresolved dependency: org.scalameta#semanticdb-scalac_2.12.6;2.1.7: not found”

PR #79, opened 2018-07-02
Workaround #77 use maven central over http (fix #77)

### LINKED ISSUE #77: Find out what we can do to release for jdk6 instead of jdk7

MasseGuillaume:


MasseGuillaume:
@julienrf all green.

.travis.yml
```diff
@@ -5,7 +5,7 @@ scala:
   - 2.12.6
   - 2.13.0-M4
 jdk:
-  - openjdk7
+  - openjdk6
```
lrytz: i think we still need the explicit `jdk_switcher` command, no? https://github.com/scala/scala-collection-compat/commit/f43e7725b57fe1688a9024bba3d6cfe0e712304d#diff-354f30a63fb0907d4ad57269548329e3L48

admin/repositories-jdk6
```diff
@@ -0,0 +1,7 @@
+[repositories]
+local
+local-preloaded-ivy: file:///${sbt.preloaded-${sbt.global.base-${user.home}/.sbt}/preloaded/}, [organization]/[module]/[revision]/[type]s/[artifact](-[classifier]).[ext]
+local-preloaded: file:///${sbt.preloaded-${sbt.global.base-${user.home}/.sbt}/preloaded/}
+maven-central-over-http: http://repo1.maven.org/maven2/
```
lrytz: so there's also an issue with downloading from maven central, which is worked around here? this doesn't fix publishing to sonatype, right? could you explain in the commit message?

MasseGuillaume: Correct, we should also change the publishTo to http.

https://github.com/scala/sbt-scala-module/blob/5c7c76c003cf681ca8d3199bec4cc8b8477942eb/src/main/scala/ScalaModulePlugin.scala#L117-L119

It's using:

```
> show compat/publishTo
[info] Some(sonatype-snapshots: https://oss.sonatype.org/content/repositories/snapshots)
```

However, this would be a security vulnerability. I will close this PR. The other solution to use jdk8 for sbt but use annother jdk forking makes sense.

PR #88, opened 2018-07-05
Run sbt with jdk8 and test with the jdk from the build matrix (fix #77)

### LINKED ISSUE #77: Find out what we can do to release for jdk6 instead of jdk7

MasseGuillaume:


MasseGuillaume:
travis-ci is not picking up the changes for some reason.

PR #91, opened 2018-07-11
Add helper traits to provide 2.13 syntax for 2.12 collections

MasseGuillaume:
For example:

build.sbt

```scala
unmanagedSourceDirectories in Compile += {
  val sourceDir = (sourceDirectory in Compile).value
  CrossVersion.partialVersion(scalaVersion.value) match {
    case Some((2, n)) if n >= 13 => sourceDir / "scala-2.13+"
    case _                       => sourceDir / "scala-2.13-"
  }
}
```

`src/main/scala/MyGrowable.scala`

```scala
import scala.collection.mutable

class MyGrowable[A](a: mutable.Set[A]) extends MyGrowableTemplate[A] {
  def addOne(elem: A): this.type = a += elem
}

```

`src/main/scala-2.13+/CompatMyGrowable.scala`

```scala
import scala.collection.mutable

trait MyGrowableTemplate[A] extends mutable.Growable[A] // addOne / + are already available
```

`src/main/scala-2.13-/CompatMyGrowable.scala`

```scala
import scala.collection.{generic, compat}

trait MyGrowableTemplate[A] extends generic.Growable[A] with compat.GrowableProxy[A]
```

### DIFF
--- a/compat/src/main/scala-2.11_2.12/scala/collection/compat/Proxy.scala
+++ b/compat/src/main/scala-2.11_2.12/scala/collection/compat/Proxy.scala
@@ -1,0 +1,47 @@
+package scala.collection.compat
+
+import scala.collection.GenTraversableLike
+import scala.collection.generic
+import scala.collection.{immutable => cimmutable}
+
+// Provide 2.13 syntax for 2.12 collections
+
+trait GrowableProxy[-A] { _: generic.Growable[A] =>
+  def addOne(elem: A): this.type
+  @`inline` override final def += (elem: A): this.type = addOne(elem)
+}
+
+trait ShrinkableProxy[-A] { _: generic.Shrinkable[A] =>
+  def subtractOne(elem: A): this.type
+  @`inline` override final def -= (elem: A): this.type = subtractOne(elem)
+}
+
+trait ClassNameProxy { _: GenTraversableLike[_, _] =>
+  def className: String
+  override def stringPrefix: String = className
+}
+
+trait ImmutableSetProxy[A] { _: cimmutable.Set[A] =>
+  def excl(elem: A): cimmutable.Set[A]
+  @`inline` final def - (elem: A): cimmutable.Set[A] = excl(elem)
+
+  def incl(elem: A): cimmutable.Set[A]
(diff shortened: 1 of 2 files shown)


compat/src/main/scala-2.11_2.12/scala/collection/compat/Proxy.scala
```diff
@@ -0,0 +1,47 @@
...
+trait ImmutableMapProxy[A, B] { _: cimmutable.Map[A, B] =>
+  def remove(key: A): cimmutable.Map[A, B]
+  @`inline` final def - (key: A): cimmutable.Map[A, B] = remove(key)
+
+  def updated[B1 >: B](key: A, value: B1): cimmutable.Map[A, B1]
+  override def + [B1 >: B](kv: (A, B1)): cimmutable.Map[A, B1] = updated(kv._1, kv._2)
+}
+
+trait MapOps[A, B] {
+  def filterInPlace(p: ((A, B)) => Boolean): this.type
```
julienrf: I’m not sure this is related to this PR

julienrf:
Another way to solve this problem would be to change 2.13’s collections to make the alphabetic name (e.g. `addOne`) an alias of the symbolic operator (e.g. `+=`) instead of the other way around.

MasseGuillaume:
> Another way to solve this problem would be to change 2.13’s collections to make the alphabetic methods (e.g. addOne) aliases of the symbolic operators (e.g. +=) instead of the other way around.

Indeed, let's do this instead.

MasseGuillaume:
@julienrf I would prefer we put this in scala/scala.

harpocrates:
> Another way to solve this problem would be to change 2.13’s collections to make the alphabetic methods (e.g. `addOne`) aliases of the symbolic operators (e.g. `+=`) instead of the other way around.

It doesn't seem that 2.13 went in that direction. In light of that, would it be worth reconsidering this PR?

julienrf:
Hey @harpocrates, this PR is out of the scope if the `scala-collection-compat` project. Please create a separate project with these helper traits if you need them.

PR #97, opened 2018-07-16
Make sure foo.to[X] has trailing brackets

MasseGuillaume:


julienrf:
Do you have any idea where the CI failure comes from?

PR #98, opened 2018-07-17
Don't use bundle to publish scalafix rules

MasseGuillaume:
This PR is to make my scalafix hack possible: https://github.com/MasseGuillaume/scalafix/commit/4108726c5239a584c440753fea9152b6d7467667

It's not possible to use bundles to fetch the migration rule for some reason. It's probably going to be useful when we have scalafix 0.6.X.


MasseGuillaume:
@julienrf can you merge this one?

Issue #99, opened 2018-07-17
Rewrite scala.collection.LinearSeq => scala.collection.immutable.List is incorrect


MasseGuillaume:
The Subclass of LinearSeq are: `List`, `Stream`, `Queue`, `Stack`.

It would break with the following call at use site:

```scala
class A(v: LinearSeq[Int])

new A(Stream(1))
```

becomes

```scala
class A(v: List[Int])

new A(Stream(1))
// found Steam[Int]
// requires List[Int]
```

julienrf:
We should indeed rewrite only `LinearSeq.apply(...)` to `List.apply(...)`. Occurrences of `LinearSeq` in type position should not be changed. If we want to not change how types are inferred, we could rewrite to `List.apply(...): LinearSeq[X]` instead.

MasseGuillaume:
It Looks like LinearSeq was not removed after all: https://github.com/scala/scala/blob/2.13.x/src/library/scala/collection/LinearSeq.scala

Can we just remove the rewrite rule?

julienrf:
Yes.

PR #100, opened 2018-07-17
Fixes for akka rewrite

### LINKED ISSUE #87: CrossCompat add compat import only if a symbol is replaced.

MasseGuillaume:
fix #87

With those changes, we can run the CrossCompat on Akka and compile the code without warning/errors and pass all the test.

Diff here:
https://github.com/MasseGuillaume/akka/commit/279fa620868fc80685a64951109bb2499af8f27b


scalafix/input/src/main/scala/fix/LinearSeqSrc.scala
```diff
@@ -1,10 +0,0 @@
-/*
-rule = "scala:fix.CrossCompat"
- */
-package fix
-
-import scala.collection.LinearSeq
-
-object LinearSeqSrc {
-  LinearSeq(1)
-}
```
julienrf: We should still support this case

scalafix/rules/src/main/scala/fix/Stable212Base.scala
```diff
@@ -45,17 +45,21 @@ trait Stable212Base extends CrossCompatibility { self: SemanticRule =>
   val traversable = exact(
     "_root_.scala.package.Traversable#",
     "_root_.scala.collection.Traversable#",
-    "_root_.scala.package.Iterable#",
-    "_root_.scala.collection.Iterable#"
```
julienrf: Why do we remove these?

scalafix/rules/src/main/scala/fix/Stable212Base.scala
```diff
@@ -66,11 +70,6 @@ trait Stable212Base extends CrossCompatibility { self: SemanticRule =>
         "scala.collection.Traversable" -> "scala.collection.Iterable"
       )
 
-    val linearSeqToList =
-      ctx.replaceSymbols(
-        "scala.collection.LinearSeq" -> "scala.collection.immutable.List",
-      )
```
julienrf: Why remove this rule?

scalafix/input/src/main/scala/fix/LinearSeqSrc.scala
MasseGuillaume: https://github.com/scala/scala-collection-compat/issues/99#issue-341891540

scalafix/rules/src/main/scala/fix/Stable212Base.scala
MasseGuillaume: we use this to detect if we need to add the compat import for 

```scala
     ctx.replaceSymbols(
        "scala.Traversable"            -> "scala.Iterable",
        "scala.collection.Traversable" -> "scala.collection.Iterable"
      )
```

It was incorrectly adding compat import.

scalafix/rules/src/main/scala/fix/Stable212Base.scala
MasseGuillaume: https://github.com/scala/scala-collection-compat/issues/99#issue-341891540

scalafix/input/src/main/scala/fix/LinearSeqSrc.scala
MasseGuillaume: No, if it's incorrect we should remove it. I'm using it to migrate akka, if we keep this rewrite in It's not possible to use scala-collection-compat to rewrite code.

PR #104, opened 2018-07-20
Add missing .to[CC] base classes

MasseGuillaume:


julienrf:
It seems that there is still a classpath issue :(

PR #107, opened 2018-07-20
Rewrite for strict reverseMap and filterKeys (fix #85)

### LINKED ISSUE #85: Rewrite for reverseMap, filterKeys
Similar to `mapValues`  https://github.com/scala/scala-collection-compat/pull/60

* [x] `reverseMap`
* [ ] `filterKeys` 

now return a view, where before it was a strict collection.

MasseGuillaume:


scalafix/input/src/main/scala/fix/Roughly212Src.scala
```diff
@@ -0,0 +1,45 @@
...
+  val d = 1 -> 1
+  val d2 = 1L -> 1
+  val multi = new m.HashMap[Int, m.Set[Int]] with m.MultiMap[Int, Int]
+
+  // i.SortedMap(d).mapValues(id): i.SortedMap[Int, Int]
+  // m.SortedMap(d).mapValues(id): c.SortedMap[Int, Int]
+  i.IntMap(d).mapValues(f2)      : i.Map[Int, Int]
+  i.LongMap(d2).mapValues(f2)    : i.Map[Long, Int]
+  i.Map(d).mapValues(id)         : i.Map[Int, Int]
+  m.LongMap(d2).mapValues(f2)    : c.Map[Long, Int]
```
julienrf: I think users would typically expect a `m.LongMap[Int]` here, instead of upcasting to `Map[Long, Int]`. If we rewrite `m.LongMap(d2).mapValues(f2)` to `m.LongMap(d2).mapValues(f2).to(LongMap)`, does that still cross-compile?

MasseGuillaume: this is the test input, I'm ascribing to make sure we preserve the same output type.

MasseGuillaume: It's typed this way in 2.12:

https://static.javadoc.io/org.scala-lang/scala-library/2.12.6/scala/collection/immutable/LongMap.html#filterKeys(p:K=%3EBoolean):scala.collection.immutable.Map[K,V]

julienrf: Ah, indeed, the return type is `Map`. I thought that operation was overridden to refine the return type but that’s not the case. Alright, then!

PR #108, opened 2018-07-20
Rewrite xs.to => xs.to(CC) (fix #106)

### LINKED ISSUE #106: unsupported .to[CC] pattern
```scala
def f(xs: List[Int]): Unit = ()
f(Set(1).to)
```

It should insert `.to(List)`

MasseGuillaume:
For example: https://github.com/akka/akka/blob/9fae4e283947c956e43aa37f83156f0e728b5965/akka-actor/src/main/scala/akka/io/Udp.scala#L282-L283

### DIFF
--- a/scalafix/rules/src/main/scala/fix/Stable212Base.scala
+++ b/scalafix/rules/src/main/scala/fix/Stable212Base.scala
@@ -185,6 +185,32 @@ trait Stable212Base extends CrossCompatibility { self: SemanticRule =>
     else Patch.empty
   }
 
+   def extractCollection(toCol: Tree): String = {
+     toCol match {
+       case Term.ApplyType(q"scala.Predef.fallbackStringCanBuildFrom", _) =>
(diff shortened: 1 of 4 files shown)


scalafix/rules/src/main/scala/fix/Stable212Base.scala
```diff
@@ -185,6 +185,32 @@ trait Stable212Base extends CrossCompatibility { self: SemanticRule =>
...
+         throw new Exception(
+           s"""|cannot extract collection from .to
+               |
+               |---------------------------------------------
+               |syntax:
+               |${toCol.syntax}
+               |
+               |---------------------------------------------
+               |structure:
+               |${toCol.structure}""".stripMargin
```
julienrf: Should we standardize the layout of error messages?

MasseGuillaume: This is from the breakout PR, I'm just going to reuse this function.

PR #110, opened 2018-07-20
Make sure .to rewrite is idempotent

MasseGuillaume:


MasseGuillaume:
Yeah or if you did some part manually, not to apply it twice.

PR #111, opened 2018-07-21
Rewrite Future.{onFailure, onSuccess} => onComplete (fix #94)

### LINKED ISSUE #94: migrate scala 2.12 @deprecated
Deprecated methods in 2.12, removed in 2.13:

```scala
[ ] scala.concurrent

  Future

    @deprecated("use `foreach` or `onComplete` instead (keep in mind that they take total rather than partial functions)", "2.12.0")
    def onSuccess[U](pf: PartialFunction[T, U])(implicit executor: ExecutionContext): Unit

    @deprecated("use `onComplete` or `failed.foreach` instead (keep in mind that they take total rather than partial functions)", "2.12.0")
    def onFailure[U](@deprecatedName('callback) pf: PartialFunction[Throwable, U])(implicit executor: ExecutionContext): Unit

    @deprecated("use the overloaded version of this method that takes a scala.collection.immutable.Iterable instead", "2.12.0")
    def find[T](@deprecatedName('futurestravonce) futures: TraversableOnce[Future[T]])(@deprecatedName('predicate) p: T => Boolean)(implicit executor: ExecutionContext): Future[Option[T]] 

    @deprecated("use Future.reduceLeft instead", "2.12.0")
    def reduce[T, R >: T](futures: TraversableOnce[Future[T]])(op: (R, T) => R)(implicit executor: ExecutionContext): Future[R] = {

```

MasseGuillaume:


scalafix/output212/src/main/scala/fix/FutureSrc.scala
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  fs.onComplete {
+    case scala.util.Success(x) if x > 0 => println("x > 0")
+    case scala.util.Success(x) if x < 0 => println("x < 0")
+    case _ => ()
+  }(ec)
+
+  fs.onComplete{
+    case scala.util.Failure(e) => pf.applyOrElse(e, Predef.identity[Throwable])
```
MasseGuillaume: This is unsafe since `pf` could take another `e` from the scope.

scalafix/input/src/main/scala/fix/FutureSrc.scala
```diff
@@ -0,0 +1,22 @@
...
+import scala.concurrent.{Future, ExecutionContext}
+import java.lang.Throwable
+
+class FutureSrc(fs: Future[Int])(implicit ec: ExecutionContext){
+  class E1 extends Throwable
+  class E2 extends Throwable
+
+  fs.onFailure {
+    case _: E1 => println("E1")
+    case _: E2 => println("E2")
```
julienrf: Can you also try with a different pattern? For instance a constructor pattern.

MasseGuillaume: I gave it a shot with function application (`fs.onSuccess(pf)`), but it's not possible since the rewrite would be:

```scala
fs.onComplete {
  case scala.util.Success(v) => pf.applyOrElse(v, Predef.identity[TYPE_OF_FS])
  case _ => ()
}
```

we don't know the type: `TYPE_OF_FS` since it could come from an expression.

Do you have other patterns in mind? Can you give me a code example for the `constructor pattern` ?

MasseGuillaume:
I agree with you it's not really in scope. However `Future`s are a central piece of code of your everyday Scala. In the Akka codebase, for example, this migration rule would be really useful.

julienrf:
I agree that it’s something useful to fix warnings of projects compiled with 2.12. But I think it should have a lower priority than incompatibilities that had no warning in 2.12.

scalafix/input/src/main/scala/fix/FutureSrc.scala
julienrf:
~~~ scala
fs.onFailure {
  case _: E1 => println("E1")
  case E2(_) => println("E2")
  case E3 => println("E3")
}
~~~

MasseGuillaume: sure, I can add a test.

MasseGuillaume:
@julienrf rebased

PR #115, opened 2018-07-24
Preserve binary compatibility for Factory

smarter:
Replace the trait Factory by an abstract type upper-bounded by
CanBuildFrom, this way usages of CanBuildFrom can be replaced by Factory
without breaking binary compatibility.

This has the unfortunate consequence that all methods available on
CanBuildFrom are still available on Factory when compiling under old
Scala versions, but that seems like a price worth paying for BC.


smarter:
Also /cc king-of-binary-compatibility @sjrd to tell me if that makes sense :)

sjrd:
AFAICT, yes, it makes sense. A confirmation from MiMa would be good.

julienrf:
> This has the unfortunate consequence that all methods available on
> CanBuildFrom are still available on Factory when compiling under old
> Scala versions, but that seems like a price worth paying for BC.

I think that’s fine because this code will be used to cross-compile with 2.13, and on 2.13 the `CanBuildFrom` API will raise compilation error if it used on a `Factory` instance.

MasseGuillaume:
I will add a follow-up PR to add mima as a unit test.

This is how you can call mima as a library: https://gist.github.com/MasseGuillaume/8b4ac3a7ec4e0d9fb9be0fb2e03b200c#file-cli-scala-L1

I will create two sbt projects then use BuildInfo to make their classpath available to the unit test.

MasseGuillaume:
@smarter Here you go: https://github.com/MasseGuillaume/scala-collection-compat/commit/945d1c8ed4463a1f5086514f01d6d88f61589e8e

You can cherry-pick this commit and add some tests.

smarter:
@MasseGuillaume Thanks, done!

smarter:
Maybe something needs to be done to run the added tests on the CI too?

MasseGuillaume:
https://github.com/MasseGuillaume/scala-collection-compat/commit/fecc6b508f64678fd3f99d5a26577a59356d19a4

MasseGuillaume:
Good to go: https://travis-ci.org/scala/scala-collection-compat/jobs/408488304#L1051

julienrf:
Awesome guys!

compat/src/main/scala-2.11_2.12/scala/collection/compat/PackageShared.scala
```diff
@@ -10,6 +10,33 @@ import scala.{collection => c}
 private[compat] trait PackageShared {
   import CompatImpl._
 
+  /**
+    * A factory that builds a collection of type `C` with elements of type `A`.
+    *
+    * @tparam A Type of elements (e.g. `Int`, `Boolean`, etc.)
+    * @tparam C Type of collection (e.g. `List[Int]`, `TreeMap[Int, String]`, etc.)
+    */
+  type Factory[-A, +C] <: CanBuildFrom[Nothing, A, C] // Ideally, this would be an opaque type
```
xuwei-k: https://github.com/scala/scala-collection-compat/pull/152

SethTisue:
a new chapter in this story: #275

PR #117, opened 2018-07-26
Cross compile scalafix input and output with 2.11, 2.12 and 2.13

MasseGuillaume:
With this PR I make sure the CrossCompat rules does compile in 2.11, 2.12 and 2.13. This is a bit envolve since we need to have a static sbt build (we cannot set the scalaVersion with crossScalaVersion). Another difficulty is to make sure all the test are discovered with scalafix-testkit. Finally, we need to update the ci to match the new static build.


MasseGuillaume:
@julienrf good to go. This should unblock me for SortedMap.unsorted.

project/Version.scala
```diff
@@ -0,0 +1,18 @@
+case class Version(major: Int, minor: Int, patch: Int) {
```
MasseGuillaume: @julienrf alternatively, we could use scalacenter/versions if you merge https://github.com/scalacenter/versions/pull/3 and push a tag.

scalafix/input/src/main/scala/fix/BreakoutSrc.scala
```diff
@@ -58,7 +58,7 @@ class BreakoutSrc(ts: Traversable[Int], vec: Vector[Int], list: List[Int], seq:
   (ts ++ ts )(breakOut): Set[Int]
 
   // `TraversableLike ++:`
-  (ts ++: ts)(breakOut): Set[Int]
+  // (ts ++: ts)(breakOut): Set[Int]
```
MasseGuillaume: I have to figure out this one, the rewritten output does not cross-compile.

scalafix/input/src/main/scala/fix/BreakoutSrc.scala
```diff
@@ -97,10 +97,8 @@ class BreakoutSrc(ts: Traversable[Int], vec: Vector[Int], list: List[Int], seq:
   List(1 -> "1").map(x => x)(breakOut): immutable.HashMap[Int, String]
   List(1 -> "1").map(x => x)(breakOut): immutable.ListMap[Int, String]
   List(1 -> "1").map(x => x)(breakOut): immutable.TreeMap[Int, String]
-  List(1 -> "1").map(x => x)(breakOut): mutable.SortedMap[Int, String]
```
MasseGuillaume: mutable.SortedMap is only available in 2.12+

scalafix/output/src/main/scala/fix/BreakoutSrc.scala
```diff
@@ -58,7 +58,7 @@ class BreakoutSrc(ts: Iterable[Int], vec: Vector[Int], list: List[Int], seq: Seq
   (ts.iterator ++ ts ).to(scala.collection.immutable.Set): Set[Int]
 
   // `TraversableLike ++:`
-  (ts ++: ts.view).to(scala.collection.immutable.Set): Set[Int]
+  // (ts ++: ts)(breakOut): Set[Int]
```
MasseGuillaume: This does not compile in 2.13.0-M4.

```scala
value ++: is not a member of scala.collection.View[Int]
```
I think it will be solved via this PR: https://github.com/scala/scala/pull/6674

julienrf: I see, that’s because we don’t have `++:` on `Iterable`.

MasseGuillaume:
@julienrf can you merge this PR? Thanks.

PR #120, opened 2018-08-01
Script to scrape the latest Scala 2.13 version from artifactory

MasseGuillaume:


project/LatestScala213.scala
```diff
@@ -0,0 +1,39 @@
...
+import com.github.nscala_time.time.Imports._
+
+import java.nio.file._
+import java.net.URL
+
+// Fetch the latest Scala 2.13 by scraping the artifactory web directory
+// We just search the latest version by date
+// run the command: latest-213 in sbt
+// NB. maven-metadata.xml does not point to the latest version
+object LatestScala {
```
julienrf: When is this object initialized? I prefer if it’s only initialized if the `latest-213` command is run so that we can load the project even if we are offline.

Also, I think it would be better to not cache the result so that if I run the `latest-213` command twice I really get the latest at the time I run the command.

MasseGuillaume: It's only initialized once when you call the `latest-213` command. I will change the val to a def so you can rerun and get the latest output.

MasseGuillaume: Done.

PR #121, opened 2018-08-01
Simplify cross-compilation for scalafix output

MasseGuillaume:

PR #122, opened 2018-08-01
 Implement experimental rule for unordering operation on Ordered collection

MasseGuillaume:
Depends on: https://github.com/scala/scala-collection-compat/pull/121
Related to: https://github.com/scala/bug/issues/11037

Summary:

map/flatMap behave differently on Ordered collection in 2.13 vs 2.12:

In Scala 2.12, if there is no Ordering when doing map or flatMap, it will convert the collection to the unordered one:

```scala
val bitset = collection.BitSet(1)

// with ordering
bitset.map(x => x) // collection.BitSet(1)

// without ordering
case class Unordered(v: Int)
bitset.map(x => Unordered(x)) // Set(Unordered(1))
```

In Scala 2.13, it's a compilation error:

```scala
bitset.map(x => Unordered(x)) // Set(Unordered(1))
          ^
error: No implicit Ordering defined for Unordered.
```

It's in the Experimental rules because we need the type of the collection (`bitset` in our example).


MasseGuillaume:
@julienrf we should fix: https://github.com/scala/bug/issues/11037. Let's merge this PR and create a followup PR to change it back to unsorted.

julienrf:
The issue has been fixed in Scala.

MasseGuillaume:
@julienrf I'm not sure this was fixed. I only modified the `.unsorted` method to return the most precise collection. We still have the case where SortedMap.map(unOrdered) does not compile in 2.13.

Also, I forgot in this PR to handle for comprehensions.

julienrf:
> We still have the case where SortedMap.map(unOrdered) does not compile in 2.13

The cross-compatible way to achieve that is:

~~~ scala
import scala.collection.compat._
SortedMap(1 -> 2).unsorted.map(toUnordered)
~~~

build.sbt
```diff
@@ -33,13 +34,16 @@ lazy val scala211   = "2.11.12"
 lazy val scala212   = "2.12.6"
 lazy val scalaJs213 = "2.13.0-M4" // Scala.js does no have -pre
 
-lazy val scala213 = "2.13.0-M4"
-// lazy val scala213 = "2.13.0-pre-3ae6282" // use the sbt command `latest-213` to fetch the latest version
+lazy val scala213    = "2.13.0-M4"
+lazy val scala213Pre = LatestScala.getLatestScala213()
```
MasseGuillaume: This is a both a blessing and a curse. It's good since it can allow us to catch bugs early, It's bad since it's makes things more fragile.

build.sbt
```diff
@@ -176,45 +192,74 @@ lazy val `scalafix-input` = project
...
+  "scalafix/output",
+  _.settings(sharedScalafixSettings)
+    .settings(dontPublish)
+    .disablePlugins(ScalafixPlugin)
+)
 
-lazy val output212Plus =
-  Def.setting((baseDirectory in ThisBuild).value / "scalafix/output212+/src/main")
-lazy val addOutput212Plus = unmanagedSourceDirectories in Compile += output212Plus.value / "scala"
+val sources = Map( // scala211   scala212   scala213   scala213Pre
```
MasseGuillaume: Since .unsorted is only available in 2.13.0-pre-... I need a way to make the output compile with a complex build matrix. I tought it would make more sense to just remove the directory structure (output212, output212Plus, output213) and make this more fine-grained, per-file.

MasseGuillaume:
@julienrf This PR is ready.

PR #129, opened 2018-08-02
Format code with scalafmt + test format via travis

MasseGuillaume:


admin/scalafmt.sh
```diff
@@ -0,0 +1,20 @@
...
+  curl -L -o $COURSIER https://git.io/vgvpD
+  chmod +x $COURSIER
+fi
+
+if [ ! -f $SCALAFMT ]; then
+  $COURSIER bootstrap com.geirsson:scalafmt-cli_2.11:$VERSION --main org.scalafmt.cli.Cli -o $SCALAFMT
+  chmod +x $SCALAFMT
+fi
+
+$SCALAFMT "$@"
```
julienrf: Why is it useful to run scalafmt on the CI?

MasseGuillaume: It's to make sure you don't push code that is not formatted. I create a separate matrix entry so you know that the CI failed only because the code was not formatted, and not more serious test failures.

julienrf:
How to legitimately agree on the code style to use? For instance, I prefer this style of scaladoc comments:

~~~
/**
  * Description
  */
~~~

over this one:

~~~
/**
 * Description
 */
~~~

MasseGuillaume:
I manually formatted code:

before:
https://github.com/scala/scala-collection-compat/blob/v0.1.1/src/main/scala-2.11_2.12/scala/collection/compat/package.scala#L14

after:
https://github.com/scala/scala-collection-compat/blob/93d3d663fef315e645cb010efadfc618a062554c/compat/src/main/scala-2.11_2.12/scala/collection/compat/PackageShared.scala#L44

PR #143, opened 2018-08-07
Add TraversableOnce.knownSize

MasseGuillaume:


MasseGuillaume:
Let's put this PR on hold: https://github.com/scala/bug/issues/11065. It looks like `knownSize` is not correctly implemented in 2.13.

PR #149, opened 2018-09-04
Fix release process

MasseGuillaume:


julienrf:
Deployment looks good so far, thanks! https://travis-ci.org/scala/scala-collection-compat/builds/424383440

julienrf:
Hmm, that was too early to say victory :)
It seems that the CI command tries to `publishSigned` the `binary-compat` project.

PR #151, opened 2018-09-06
Add AnyTap, AnyPipe and Using

martijnhoekstra:
To incite discussion.

2.13.x has nice new library additions, and it would be nice if we can offer source-compatibility beyond collections.

This cherry-picks `AnyTap`, `AnyPipe` and `Using` from 2.13 to compat.

### DIFF
--- a/compat/src/main/scala-2.11_2.12/scala/util/compat/ChainingOps.scala
+++ b/compat/src/main/scala-2.11_2.12/scala/util/compat/ChainingOps.scala
@@ -1,5 +1,4 @@
-package scala
-package util
+package scala.util.compat
 
 trait ChainingSyntax {
   implicit final def scalaUtilChainingOps[A](a: A): ChainingOps[A] = new ChainingOps(a)
@@ -20,7 +19,7 @@ final class ChainingOps[A](val self: A) extends AnyVal {
    *  @tparam U     the result type of the function `f`.
    *  @return       the original value `self`.
    */
-  def tap[U](f: A => U): self.type = {
+  def tap[U](f: A => U): A = {
     f(self)
     self
   }
--- a/compat/src/main/scala-2.11_2.12/scala/util/compat/Using.scala
+++ b/compat/src/main/scala-2.11_2.12/scala/util/compat/Using.scala
@@ -139,7 +139,8 @@ object Using {
 
     @inline def safeAddSuppressed(t: Throwable, suppressed: Throwable): Unit = {
       // don't `addSuppressed` to something which is a `ControlThrowable`
-      if (!t.isInstanceOf[ControlThrowable]) t.addSuppressed(suppressed)
+      // nor to scalaJS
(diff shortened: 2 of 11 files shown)


martijnhoekstra:
The scala-js failures depend on https://github.com/scala-js/scala-js/commit/1fbc66e6973756381042b4d6809bb5f602ad4999 which is in master, but not published yet.

The JVM 6 error is around AutoClosable, and I'm open to suggestions to that

NthPortal:
The more I think about this, the more I think it's a good idea (though it might also suggest renaming the artifact/repo to scala-2.13-compat). These things are really useful, and resource closing in particular is non-trivial and usually done wrong.

NthPortal:
@martijnhoekstra for JVM 6, there's `java.lang.Closable` I think? I'm not sure what the tradeoffs of using that vs `j.l.AutoClosable` are

martijnhoekstra:
@NthPortal I changed to Closeable, but at this point it's not clear to me why the original used AutoCloseable -- I did nothing special to change it, but if that's universally correct, why not have it on Closeable in 2.13.x (other than it being frozen)?

martijnhoekstra:
Actually, I don't know how to do this.

NthPortal:
@martijnhoekstra because `AutoCloseable` is (a) what try-with-resources works on, and (b) a parent of `Closeable`, so it is strictly more general

julienrf:
I agree that this would be good to extend the scope of scala-collection-compat to other parts of the scala-library. I’m not sure who is going to check that what we have in the compat library will stay consistent with the standard library and who is going to maintain the compat library in general?

martijnhoekstra:
@julienrf from a high-level perspective, the invariant that should hold, I assume, is if some identifier`f` exists in compat then `2.12.compat.f == 2.13.compat.f == 2.13.f`. How we can and should test that I'm unsure of.

I would sort of assume that the maintainers of this library are still going to maintain it if it expands from a collection-compat library to a general stdlib-compat library, but from your comment I get the vibe that's not a given 😬 

For the functionality I implemented in this PR, I don't understand how the exception suppression works, so I think I'll need to cut `Resource[AutoCloseable]` from this PR entirely, or I can't guarantee the bandwidth that that particular inclusion won't hold up the wider issue.

SethTisue:
see also https://github.com/bigwheel/util-backports /cc @bigwheel

bigwheel:
Thanks @SethTisue for mention to me because I didn't know this useful library.
If this PR is merged after resolving above problem, I should think about stop to maintain https://github.com/bigwheel/util-backports and put URL https://github.com/scala/scala-collection-compat to its README.md.

martijnhoekstra:
With how much is in flux around both the features, and what's happening in the stdlib with this kind of stuff, combined with my sustained low bandwidth for OS work lately, I'm really out of the loop, and won't be able to bring this to a conclusion.

Issue #137, opened 2018-08-06
No implicit for Factory[Nothing, Seq[Nothing]]


MasseGuillaume:
2.12

```scala
implicitly[CanBuildFrom[Nothing, Nothing, Seq[Nothing]]]
```

2.13

```scala
implicitly[Factory[Nothing, Seq[Nothing]]]
```

2.12 + compat
```scala
import scala.collection.compat._
implicitly[Factory[Nothing, Seq[Nothing]]]
// could not find implicit value for parameter e: collection.compat.Factory[Nothing,Seq[Nothing]]
```

Use case:
https://github.com/akka/akka/blob/a0e79a0542dc0fc98bf8431a1cfd1a454f45b282/akka-stream-tests/src/test/scala/akka/stream/scaladsl/CollectionSinkSpec.scala#L39

PR #152, opened 2018-09-12
fix the "Factory" type alias definition

xuwei-k:
avoid `import scala.collection.compat._` everywhere in Scala 2.12 and 2.11

### collection-compat v0.2.0

```
Welcome to Scala 2.12.6 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_181).

scala> implicitly[scala.collection.compat.Factory[Int, List[Int]]]
<console>:12: error: could not find implicit value for parameter e: collection.compat.Factory[Int,List[Int]]
       implicitly[scala.collection.compat.Factory[Int, List[Int]]]
                 ^

scala> import scala.collection.compat._
import scala.collection.compat._

scala> implicitly[scala.collection.compat.Factory[Int, List[Int]]]
res1: collection.compat.Factory[Int,List[Int]] = scala.collection.generic.GenTraversableFactory$$anon$1@41b12185
```

### collection-compat v0.1.1

```
Welcome to Scala 2.12.6 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_181).

scala> implicitly[scala.collection.compat.Factory[Int, List[Int]]]
res0: scala.collection.compat.Factory[Int,List[Int]] = scala.collection.compat.Factory$$anon$1@33600943
```

### DIFF
--- a/compat/src/main/scala-2.11_2.12/scala/collection/compat/PackageShared.scala
+++ b/compat/src/main/scala-2.11_2.12/scala/collection/compat/PackageShared.scala
@@ -16,7 +16,7 @@ private[compat] trait PackageShared {
    * @tparam A Type of elements (e.g. `Int`, `Boolean`, etc.)
    * @tparam C Type of collection (e.g. `List[Int]`, `TreeMap[Int, String]`, etc.)
    */
-  type Factory[-A, +C] <: CanBuildFrom[Nothing, A, C] // Ideally, this would be an opaque type
+  type Factory[-A, +C] = CanBuildFrom[Nothing, A, C] // Ideally, this would be an opaque type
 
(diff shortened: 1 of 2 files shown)


MasseGuillaume:
LGTM, Can you look if it fixes https://github.com/scala/scala-collection-compat/issues/137.

julienrf:
After thinking more about it, maybe using a type alias (as proposed in this PR) is better than using `<:`.

Indeed, with `<:`, users have to add the `import scala.collection.compat._` to get their `CanBuildFrom` instance converted to the expected `Factory` existential type. This means that libraries that use `Factory` will require users to write this additional import, whereas with a type alias that import wouldn’t be necessary.

Do other people have any thought about merging this PR?

julienrf:
Thanks @xuwei-k !

PR #163, opened 2018-10-23
fix sbt task name in migration tool doc

pjfanning:
I tried the 0.2.1 migration tool and Collection213CrossCompat appears to be the right name


julienrf:
Thanks!

PR #172, opened 2019-01-31
fix JDK 6 on CI, and bump Scala and sbt and versions

SethTisue:
modeled after @ashawley's https://github.com/scala/scala-xml/pull/271

log for a green run from that PR, for reference: https://travis-ci.org/scala/scala-xml/builds/484411533


SethTisue:
once this is green, but before I merge it, I should at least try to add `openjdk11` to the Travis-CI matrix

SethTisue:
if we want to use sbt 0.13.18 instead of 0.13.17, we may need something like https://github.com/scala/community-builds/issues/836

SethTisue:
LOL FML

```
[error] /home/travis/build/scala/scala-collection-compat/project/LatestScala213.scala:4: object file is not a member of package java.nio
[error] import java.nio.file._
[error]                 ^
[error] one error found
[error] (compile:compileIncremental) Compilation failed
```

SethTisue:
I'm just going to rip out this latest-Scala-2.13 stuff, since we're so close to RC1

SethTisue:
gah, as I get deeper into this, I'm finding that Scalafix doesn't work on Java 6, and Scalafix stuff is all through the build now, since this repo has Scalafix rules as well as the library.

let's just drop JDK 6 support instead. probably the artifacts we build will actually work on Java 6. if somebody finds they don't, they can report it and we can fix it. probably the number of people who will want to use this library is small/zero.

SethTisue:
version-bumps are now #174

Issue #180, opened 2019-02-07
roll an 0.3.0 release


SethTisue:
needed over at https://github.com/scala/community-builds/pull/849

SethTisue:
tagged, release notes attached: https://github.com/scala/scala-collection-compat/releases/tag/v0.2.2

SethTisue:
Travis publishing run: https://travis-ci.org/scala/scala-collection-compat/builds/490341153

SethTisue:
the 6 staging repos on oss.sonatype.org are orgscala-lang-1664 through orgscala-lang-1669. I hit "close" and "release" on them

SethTisue:
gah, 2.11 artifacts aren't on Maven Central

SethTisue:
haha well no wonder, wtf

```
  exclude:
    # ?                                                                                                                
    - jdk: oraclejdk8
      scala: 2.11.12
```

SethTisue:
I'm rolling 0.3.0. no code changes, but I want to verify that the full cross-publish goes smoothly, and @julienrf pointed out that although we aren't guaranteeing bincompat, 0.2.2 wasn't bincompat with 0.2.1. seems better to bump the version number to 0.3.0 to give more indication that bincompat broke.

SethTisue:
* [x] tag it, write release notes: https://github.com/scala/scala-collection-compat/releases/tag/v0.3.0
* [ ] wait for Travis to publish https://travis-ci.org/scala/scala-collection-compat/builds/490694949
* [ ] close and release staging repos
* [ ] wait for artifacts to show up:
  * https://repo1.maven.org/maven2/org/scala-lang/modules/scala-collection-compat_2.11/0.3.0/
  * https://repo1.maven.org/maven2/org/scala-lang/modules/scala-collection-compat_2.12/0.3.0/
  * https://repo1.maven.org/maven2/org/scala-lang/modules/scala-collection-compat_2.13.0-M5/0.3.0/

SethTisue:
gah, what a mess. I fixed the JDK 11 jobs to be green, but now that they're green which is good, they're also uploading artifacts to the staging repos, which we don't want. 😖

SethTisue:
steps:
* [x] tag it, include release notes: https://github.com/scala/scala-collection-compat/releases/tag/v0.3.0
* [ ] wait for Travis to publish https://travis-ci.org/scala/scala-collection-compat/builds/490694949
* [ ] close and release staging repos
[...]

PR #182, opened 2019-02-08
publish releases from JDK 8 jobs only, not JDK 11

SethTisue:
context: #180


SethTisue:
followup 9d5b318c8e3782db3fe9d4d46949d03531b36cd7

PR #188, opened 2019-03-01
[DON'T MERGE] Replace lengthCompare with sizeCompare

NthPortal:
Add logic to Collection213Upgrade to replace
`Seq#lengthCompare` with `Seq#sizeCompare`.

Migration to accompany deprecation of `lengthCompare` as proposed in scala/bug#11388.


NthPortal:
do not merge until scala/bug#11388 has a resolution decided

NthPortal:
Looks like we're keeping `lengthCompare`

PR #193, opened 2019-04-05
Update for Scala 2.13.0-RC1

szeiger:
- Stubs are now built for 2.13.0-RC1 and 2.13.0-M5
- Add ScalaJS version 1.0.0-M7
- Update scala-collection-compat version to 1.0.0 and document binary
  compatibility policy in the README

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: scala
 scala:
   - 2.11.12
   - 2.12.8
-  - 2.13.0-M5
+  - 2.13.0-RC1
 jdk:
   - oraclejdk8
   - openjdk11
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -4,6 +4,7 @@ scala:
   - 2.11.12
   - 2.12.8
   - 2.13.0-M5
```
SethTisue: let's drop M5?

.travis.yml
```diff
@@ -20,6 +21,7 @@ env:
     - SCALAJS_VERSION=
     - SCALAJS_VERSION=0.6.26
```
SethTisue: should be 0.6.27

.travis.yml
```diff
@@ -20,6 +21,7 @@ env:
...
     - SCALAJS_VERSION=1.0.0-M6
```
SethTisue: let's drop M6

README.md
```diff
@@ -8,14 +8,14 @@ This library provides some of the new APIs from Scala 2.13 to Scala 2.11 and 2.1
 To use this library, add the following to your build.sbt:
 
 ```
-libraryDependencies += "org.scala-lang.modules" %% "scala-collection-compat" % "0.2.2"
+libraryDependencies += "org.scala-lang.modules" %% "scala-collection-compat" % "1.0.0"
 ```
 
-Version 0.2.2 is compatible with Scala 2.13.0-M5. For Scala 2.13.0-M4 you should use version 0.1.1.
+Version 1.0.0 is compatible with Scala 2.13.0-RC1 and 2.13.0-M5. For Scala 2.13.0-M4 you should use version 0.1.1.
```
SethTisue: drop M5?

and I think any mention of M4 can be omitted

README.md
```diff
@@ -8,14 +8,14 @@ This library provides some of the new APIs from Scala 2.13 to Scala 2.11 and 2.1
...
+libraryDependencies += "org.scala-lang.modules" %% "scala-collection-compat" % "1.0.0"
 ```
 
-Version 0.2.2 is compatible with Scala 2.13.0-M5. For Scala 2.13.0-M4 you should use version 0.1.1.
+Version 1.0.0 is compatible with Scala 2.13.0-RC1 and 2.13.0-M5. For Scala 2.13.0-M4 you should use version 0.1.1. 
 
 Note that there are multiple ways to cross-build projects, see https://github.com/scala/collection-strawman/wiki/FAQ#how-do-i-cross-build-my-project-against-scala-212-and-scala-213.
 
-Also note that this library has not fully stabilized yet. We expect that new, binary incompatible releases of this library will be published (for 2.11, 2.12) until Scala 2.13 is getting close to its final state. Therefore you might want to avoid adding a dependency on that library to your 2.11 / 2.12 artifacts for the time being.
+Starting with version 1.0.0 backwards binary compatibility is enforced within each major version (i.e. anything prior to 2.0.0 will be compatible).
```
szeiger: Should we mention the possibility of a 2.0.0 at all or simply say that all new versions will be backwards compatible? If we get into a situation where we want to break binary compatibility, it's probably best to either resist that urge or publish under a new artifact name. I put the 2.0.0 limit in just in case.

SethTisue: I think "within each major version" would be sufficient, but no strong feeling either way.

szeiger:
The build failed because the builds script cannot parse an RC version number. It can parse M versions but discards anything after the M. If it works for M, the same should work for RC. But why are we discarding this information in `Version.parse`? Isn't it essentials for pretty much any place where you want to use a version number?

szeiger:
After a further look: The version number is used for a number of different versions, only one of which (the Scala version) will realistically be an M or RC. And in this case it is only used for a log message and the directory name used for staging artifacts. Using the verbatim version string for these things would be better but everything should work just fine the way it is now.

PR #198, opened 2019-04-19
fix travis-ci cache settings

xuwei-k:


SethTisue:
ah, good catch

(the CI failures are unrelated; `openjdk11` isn't working anymore :-/)

PR #200, opened 2019-04-23
Upgrade BuildFrom to the latest definition from 2.13

szeiger:


compat/src/main/scala-2.11_2.12/scala/collection/compat/BuildFrom.scala
```diff
@@ -41,7 +39,7 @@ object BuildFrom {
   implicit def fromCanBuildFrom[From, A, C](
       implicit cbf: CanBuildFrom[From, A, C]): BuildFrom[From, A, C] =
     new BuildFrom[From, A, C] {
-      def fromSpecificIterable(from: From)(it: Iterable[A]): C = (cbf(from) ++= it).result()
+      def fromSpecific(from: From)(it: IterableOnce[A]): C = (cbf(from) ++= it).result()
```
eed3si9n: Are there any test coverage for this change?

szeiger: I've added a test to ensure that a `BuildFrom` implementation compiles on all versions.

szeiger:
I ran a test of Slick with this PR. As expected no code changes were necessary and everything worked fine. The old code was based on `CanBuildFrom` so it only uses the strict portions of `BuildFrom` for now, and it only uses provided instances but doesn't have to implement `BuildFrom` itself. It *does* implement `Factory`, which requires different source codes for 2.11/2.11 and 2.13, but that's unavoidable if we want to keep `Factory` as a simple type alias for `CanBuildFrom` instead of providing another class for it.

PR #207, opened 2019-05-24
Collections213 => Collection213

### LINKED ISSUE #194: README.md example not working?
Two issues:

1) I couldn't get the `scalacOptions` examples for `Collections213CrossCompat` to work in https://github.com/scala/scala-collection-compat/blob/master/README.md to work, I had to explicitly `addCompilerPlugin(scalafixSemanticdb), scalacOptions += "-Yrangepos"`

2) The scala-fix version should be updated to 0.9.4

Tangentially related, I also couldn't get it to work with `2.13-RC1` because it couldn't find the semantic db to be available, so I had to make sure my scala version was set to `2.12` I'm not sure if that is intended/etc.  But some clarification would certainly be helpful!

yufangong:
Fixes #194

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ The migration rules use scalafix. Please see the [official installation instruct
 ```scala
 // project/plugins.sbt
 
-addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.9.0")
+addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.9.5")
 ```
 
 ### Collection213Upgrade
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -44,7 +44,7 @@ The migration rules use scalafix. Please see the [official installation instruct
 ```scala
 // project/plugins.sbt
 
-addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.9.0")
+addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.9.5")
```
NthPortal: I'd like someone other than me to confirm this change. I can't imagine any reason it would be bad, but I don't know much about scalafix.

Issue #210, opened 2019-05-29
publish for Scala 2.13.0-RC3


SethTisue:
PR #211

SethTisue:
I don't see anything in this repo that supports `v2.0.0#2.13.0-RC3#8` style publishing. @szeiger are you free to take this one? I'm working on the other modules.

SethTisue:
not sure how to handle this:

* go to v2.0.1? annoying downstream
* delete the old 2.0.0 tag, re-tag it, publish, then drop everything but the RC3 artifacts from staging before we release?

SethTisue:
(I don't know the history behind how this repo ended up with nonstandard publishing stuff. is it because it isn't just the module itself, but also the scalafix stuff?)

SethTisue:
oh, I'm behind the times, it works like this now:

```
# In order to build a previously released version against a new (binary incompatible) Scala release,
# a new commit that modifies (and prunes) the Scala versions in .travis.yml needs to be added on top
# of the existing tag. Then a new tag can be created for that commit, e.g., `v1.2.3#2.13.0-M5`.
# Everything after the `#` in the tag name is ignored.
```

SethTisue:
note that it isn't the same as the new-style modules (scala-xml, scala-parser-combinators) *or* the same as the old-style modules

dwijnand:
Just publish from local?

SethTisue:
out of the frying pan and into the fire

SethTisue:
when I wrote my previous comments, I hadn't understood yet how the scala-xml and scala-parser-combinators publishing worked. instead of controlling it entirely via the tag name, you push a commit that cuts the matrix in `.travis.yml` down to just the combinations you want published, and then you tag that commit

I will give that a try in this repo

SethTisue:
* prep work: #212 
* commit: e6612db640f6839b195a6e988cbc935fd2abba77
* tag: https://github.com/scala/scala-collection-compat/releases/tag/v2.0.0%232.13.0-RC3
* job: https://travis-ci.org/scala/scala-collection-compat/builds/539250899

SethTisue:
staging repos closed and released

now we watch https://repo1.maven.org/maven2/org/scala-lang/modules/scala-collection-compat_2.13.0-RC3/2.0.0/

SethTisue:
staging repo numbers: 1786, 1787, 1788

I was hoping we'd reach 1789, since I'm in Paris 🇫🇷 ✊

[...]

PR #212, opened 2019-05-30
allow republishing for new Scala versions

SethTisue:
by ignoring trailing `#junk` on TRAVIS_TAG

context is #210


NthPortal:
you could also make it not a capturing group

SethTisue:
> you could also make it not a capturing group

good point. #213

Issue #203, opened 2019-05-13
Add IterableView.mapValues


szeiger:
2.12:
```
scala> res0.mapValues(x=>  x^2)
res7: scala.collection.immutable.Map[String,Int] = Map(32 -> 1)
```

2.13:
```
scala> res0.mapValues(x=>  x^2).toMap
            ^
       warning: method mapValues in trait MapOps is deprecated (since 2.13.0): Use .view.mapValues(f). A future version will include a strict version of this method (for now, .view.mapValues(f).toMap).
res9: scala.collection.immutable.Map[String,Int] = Map(32 -> 1)
```

The recommended replacement does not compile in 2.12:

```
scala> Map(1 -> 2).view.mapValues(x => x*x)
<console>:12: error: value mapValues is not a member of scala.collection.IterableView[(Int, Int),scala.collection.immutable.Map[Int,Int]]
       Map(1 -> 2).view.mapValues(x => x*x)
                        ^
```

This could be remedied by adding an extension method for map-like `IterableView`s.

NthPortal:
don't we have a `MapView` it could be on?

SethTisue:
for anyone here on this ticket wondering about a workaround in the meantime, consider whether your code would still work if you replaced `.mapValues(f)` with `.mapValues{case (k, v) => (k, f(v))`. and if you do need the laziness, perhaps you can `.iterator` your way to laziness

japgolly:
Came here to report this, awesome that it's already been fixed! Any idea when the next release is coming?

SethTisue:
@lrytz another possible rewrite

PR #220, opened 2019-06-17
Add mapValues for IterableViews of Maps

szeiger:
Fixes https://github.com/scala/scala-collection-compat/issues/203


unoexperto:
What version of `scala-collection-compat` will it be in, guys ?

NthPortal:
@unoexperto it is in version 2.1.1

PR #221, opened 2019-06-19
Added AbstractBuilder & AbstractFactory cross-compatible base traits

ghik:
Currently there is no way to implement a `Builder` or `Factory` in a way that would compile in both Scala 2.12 and 2.13. This PR introduces `AbstractBuilder` and `AbstractFactory` which serve that purpose.

This PR is very draft, I just wanted to start a discussion about whether it's a good idea or not.

### DIFF
--- a/compat/src/main/scala-2.11_2.12/scala/collection/compat/AbstractBuilder.scala
+++ b/compat/src/main/scala-2.11_2.12/scala/collection/compat/AbstractBuilder.scala
@@ -1,0 +1,9 @@
+package scala.collection.compat
+
+import scala.collection.mutable
+
+trait AbstractBuilder[-Elem, +To] extends mutable.Builder[Elem, To] {
+  def addOne(elem: Elem): this.type
+
+  final def +=(elem: Elem): this.type = addOne(elem)
+}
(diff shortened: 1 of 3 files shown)


sjrd:
IME there's simply no way to implement any kind of collection in a cross-compatible way. Trying to "fix" the situation in collection-compat is only going to help one drop at a time.

So no, I don't think this is a good idea. Let's keep scala-collection-compat focused on *using* collections in a cross-compatible way, but not implementing them.

ghik:
Implementing a `Builder` or `Factory` does not necessarily mean that one implements a custom collection. For example, the very problem that I faced was that I needed some customized builders for already existing collections, including Java collections.

julienrf:
So far, the way scala-collection-compat works is by introducing the 2.13 syntax on older Scala versions. The idea is that eventually (when you don’t need to support older Scala versions anymore) you drop the `import scala.collection.compat._` and your project still compiles.

I’m not sure we should invest in defining *another* syntax that would work both on 2.13 and older versions. The solution in your case is probably to go with different source directories for Scala 2.13+ and 2.13-, as explained [here](https://docs.scala-lang.org/overviews/core/collections-migration-213.html#how-do-i-cross-build-my-project-against-scala-212-and-scala-213).

ghik:
> The idea is that eventually (when you don’t need to support older Scala versions anymore) you drop the import scala.collection.compat._ and your project still compiles.

Sure, if that's the assumption we want to keep then definitely this PR does not fit.

julienrf:
That being said, thanks for opening a PR and raising a discussion! I’m closing it for now but if that use case pops up a lot we might re-open the debate.

PR #225, opened 2019-06-23
Fix OSGI exports

### LINKED ISSUE #224: scala.jdk package isn't published
# Steps to reproduce

Using either the artifacts published for 2.1.0, or grabbing those created by running `sbt publishLocal`, list the files found in the Scala 2.12 artifact.

# Expected behaviour

The content of package `scala.jdk` should be there - it gets compiled *and* packaged.
Run the following commands to confirm:
* `sbt package`
* `unzip -l .cross/compat212/jvm/target/scala-2.12/scala-collection-compat_2.12-0.1.0-SNAPSHOT.jar`

# Observed behaviour

The content of package `scala.jdk` is not there.

NthPortal:
Fix OSGI exports so `scala/jdk` classfiles are not removed
from the jar.

Fixes #224.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -73,7 +73,7 @@ lazy val compat = MultiScalaCrossProject(JSPlatform, JVMPlatform)(
       }
     )
     .jvmSettings(
-      OsgiKeys.exportPackage := Seq(s"scala.collection.compat.*;version=${version.value}"),
+      OsgiKeys.exportPackage := Seq(s"scala.collection.compat.*;scala.jdk.*;version=${version.value}"),
       junit
     )
     .jsSettings(


NthPortal:
It's not clear to me how the version is bumped - is that just done by tagging for release?

build.sbt
```diff
@@ -73,7 +73,7 @@ lazy val compat = MultiScalaCrossProject(JSPlatform, JVMPlatform)(
       }
     )
     .jvmSettings(
-      OsgiKeys.exportPackage := Seq(s"scala.collection.compat.*;version=${version.value}"),
+      OsgiKeys.exportPackage := Seq(s"scala.collection.compat.*;scala.jdk.*;version=${version.value}"),
```
NthPortal: Is there a reason for this not to be the following?

```suggestion
      OsgiKeys.exportPackage := Seq(s"*;version=${version.value}"),
```

eed3si9n: I guess the intent of the OSGi proponent was to be explicit about what this "exports"? It might be good to use `*` until we actually do integration testing using the resulting bundle.

szeiger: It should be OK. The packages get expanded anyway.

szeiger: Actually no, it is not OK. Using `*` will export *all* packages that are visible, including all used `scala` subpackages. We have to be more precise and only specify the ones that are really part of this module.

lrytz: the module should export `scala.jdk` only in the 2.11 / 2.12 versions, something like

```
      OsgiKeys.exportPackage := {
        if (scalaVersion.value.startsWith("2.13."))
          Seq(s"scala.collection.compat.*;version=${version.value}")
        else
          Seq(s"scala.collection.compat.*;version=${version.value},scala.jdk.*;version=${version.value}")
      },
```

NthPortal:
closing in favour of #226

PR #217, opened 2019-06-14
Backport scala.jdk.CollectionConverters to 2.11 / 2.12

lrytz:
Fixes https://github.com/scala/scala-collection-compat/issues/208

PR #229, opened 2019-06-28
Set up MiMa

### LINKED ISSUE #162: Binary compatibility
The README currently states:

> We expect that new, binary incompatible releases of this library will be published (for 2.11, 2.12) until Scala 2.13 is getting close to its final state

When it is time to promise binary compatibility, we can 
* [x] replace that line with a statement around binary compatibility
* [ ] add MiMa checks to CI

dwijnand:
Closes #162

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -13,7 +13,9 @@ lazy val commonSettings = Seq(
                                                  |
                                                  |See the NOTICE file distributed with this work for
                                                  |additional information regarding copyright ownership.
-                                                 |""".stripMargin)))
+                                                 |""".stripMargin)),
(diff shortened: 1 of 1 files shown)


szeiger:
I don't see anything here that would run MiMa as part of CI. Is this still missing?

julienrf:
I think it comes from the `ScalaModulePlugin`:

https://github.com/scala/sbt-scala-module/blob/2.x/src/main/scala/ScalaModulePlugin.scala#L201-L204

dwijnand:
Yes, that's right.  And I tested it by locally reverting #217 and running `test`.

PR #237, opened 2019-07-25
to(Seq) doesn't evaluate collection when called on a Seq

### LINKED ISSUE #236: `to(Seq)` doesn't force a stream, where `to[Seq]` did
In 2.13, a `Stream#to(Seq)` returns identity. 

In 2.12, `Stream#to[Seq]` returns a `Vector`, as does `Stream#to(Seq)` with compat.

I would argue that's a progression, not a regression, but it does introduce a semantic, potentially breaking change, both in the face of serialization and termination, when the stream isn't forced.

special-casing the scalafix for `Stream#to(Seq)` to `Stream#toVector` would make this change safe and obvious. That could either be obviously wrong or obviously right, both is a win.

Arguably, the same should be done for any `CC#to(Seq)` where `Seq <: CC <: Stream`.

martijnhoekstra:
Fixes #236 

Not so much about elegance as about getting the job done.


martijnhoekstra:
obviously not forward bincompat. Is that needed?

PR #238, opened 2019-07-25
to(Seq) doesn't evaluate collection when called on a Seq

### LINKED ISSUE #236: `to(Seq)` doesn't force a stream, where `to[Seq]` did

martijnhoekstra:
Fixes #236 Not so much about elegance as about getting the job done.

### DIFF
--- a/compat/src/main/scala-2.11_2.12/scala/collection/compat/CompatImpl.scala
+++ b/compat/src/main/scala-2.11_2.12/scala/collection/compat/CompatImpl.scala
@@ -14,14 +14,13 @@ package scala.collection.compat
 
 import scala.reflect.ClassTag
 import scala.collection.generic.CanBuildFrom
-import scala.collection.mutable.Builder
 import scala.collection.{immutable => i, mutable => m}
 
 /* builder optimized for a single ++= call, which returns identity on result if possible
  * and defers to the underlying builder if not.
  */
-private final class IdentityPreservingBuilder[A, CC[X] <: TraversableOnce[X]](that: Builder[A, CC[A]])(implicit ct: ClassTag[CC[A]])
-    extends Builder[A, CC[A]] {
+private final class IdentityPreservingBuilder[A, CC[X] <: TraversableOnce[X]](that: m.Builder[A, CC[A]])(implicit ct: ClassTag[CC[A]])
+    extends m.Builder[A, CC[A]] {
 
   //invariant: ruined => (collection == null)
   var collection: CC[A] = null.asInstanceOf[CC[A]]
@@ -62,9 +61,9 @@ private final class IdentityPreservingBuilder[A, CC[X] <: TraversableOnce[X]](th
 }
 
 private[compat] object CompatImpl {
-  def simpleCBF[A, C](f: => Builder[A, C]): CanBuildFrom[Any, A, C] = new CanBuildFrom[Any, A, C] {
-    def apply(from: Any): Builder[A, C] = apply()
-    def apply(): Builder[A, C]          = f
+  def simpleCBF[A, C](f: => m.Builder[A, C]): CanBuildFrom[Any, A, C] = new CanBuildFrom[Any, A, C] {
+    def apply(from: Any): m.Builder[A, C] = apply()
+    def apply(): m.Builder[A, C]          = f
   }
 
   type ImmutableBitSetCC[X] = ({ type L[_] = i.BitSet })#L[X]
--- a/compat/src/main/scala-2.11_2.12/scala/collection/compat/CompatImpl.scala
+++ b/compat/src/main/scala-2.11_2.12/scala/collection/compat/CompatImpl.scala
@@ -12,7 +12,7 @@
 
 package scala.collection.compat
 
-import scala.collection.LinearSeq
+import scala.reflect.ClassTag
 import scala.collection.generic.CanBuildFrom
 import scala.collection.mutable.Builder
 import scala.collection.{immutable => i, mutable => m}
@@ -20,65 +20,37 @@ import scala.collection.{immutable => i, mutable => m}
 /* builder optimized for a single ++= call, which returns identity on result if possible
  * and defers to the underlying builder if not.
  */
-private final class IdentityPreservingSeqBuilder[A](that: Builder[A, Seq[A]])
-    extends Builder[A, Seq[A]] {
-  var collection: Seq[A] = null
+private final class IdentityPreservingBuilder[A, CC[X] <: TraversableOnce[X]](that: Builder[A, CC[A]])(implicit ct: ClassTag[CC[A]])
+    extends Builder[A, CC[A]] {
+  var collection: CC[A] = null.asInstanceOf[CC[A]]
   var ruined = false
 
   final override def ++=(elems: TraversableOnce[A]): this.type =
-      if(!ruined && collection == null && elems.isInstanceOf[Seq[_]]) {
-        collection = elems.asInstanceOf[Seq[A]]
-        this
-      }
-      else {
-        ruined = true
-        if (collection != null) that ++= collection
-        that ++= elems
-        collection = null
-        this
+      elems match {
+        case ct(ca) if (collection == null && !ruined) =>  {
+          collection = ca
+          this
+        }
+        case _ => {
+          ruined = true
+          if (collection != null) that ++= collection
+          that ++= elems
+          collection = null.asInstanceOf[CC[A]]
+          this
+        }
       }
 
   final def +=(elem: A): this.type = {
-    collection = null
+    collection = null.asInstanceOf[CC[A]]
     ruined = true
     that += elem
     this
   }
   final def clear(): Unit = {
-    collection = null
+    collection = null.asInstanceOf[CC[A]]
     if (ruined) that.clear()
   }
-  final def result(): Seq[A] = if(ruined) that.result() else if (collection eq null) Nil else collection
-}
-
-private final class IdentityPreservingLinearSeqBuilder[A](that: Builder[A, LinearSeq[A]]) extends Builder[A, LinearSeq[A]] {
-  var collection: LinearSeq[A] = null
-  var ruined = false
-
-  final override def ++=(elems: TraversableOnce[A]): this.type =
-      if(!ruined && collection == null && elems.isInstanceOf[LinearSeq[_]]) {
-        collection = elems.asInstanceOf[LinearSeq[A]]
-        this
-      }
-      else {
-        ruined = true
-        if (collection != null) that ++= collection
-        that ++= elems
-        collection = null
-        this
-      }
-
-  final def +=(elem: A): this.type = {
-    collection = null
-    ruined = true
-    that += elem
-    this
-  }
-  final def clear(): Unit = {
-    collection = null
-    if (ruined) that.clear()
-  }
-  final def result(): LinearSeq[A] = if(ruined) that.result() else if (collection eq null) Nil else collection
+  final def result(): CC[A] = if(ruined || (collection == null)) that.result() else collection
 }
 
 private[compat] object CompatImpl {
--- a/compat/src/main/scala-2.11_2.12/scala/collection/compat/CompatImpl.scala
+++ b/compat/src/main/scala-2.11_2.12/scala/collection/compat/CompatImpl.scala
@@ -12,6 +12,7 @@
 
 package scala.collection.compat
 
+import scala.collection.LinearSeq
 import scala.collection.generic.CanBuildFrom
 import scala.collection.mutable.Builder
 import scala.collection.{immutable => i, mutable => m}
@@ -19,7 +20,7 @@ import scala.collection.{immutable => i, mutable => m}
 /* builder optimized for a single ++= call, which returns identity on result if possible
  * and defers to the underlying builder if not.
  */
-private final class IdentityPreservingSeqBuilder[A, C <: Seq[A]](that: Builder[A, C])
+private final class IdentityPreservingSeqBuilder[A](that: Builder[A, Seq[A]])
     extends Builder[A, Seq[A]] {
   var collection: Seq[A] = null
   var ruined = false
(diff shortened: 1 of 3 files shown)


martijnhoekstra:
@NthPortal yes, I guess

martijnhoekstra:
I duplicated the logic for `LinearSeq`. The lack of elegance I was referring to earlier is starting to stick out even further.

compat/src/main/scala-2.11_2.12/scala/collection/compat/CompatImpl.scala
```diff
@@ -12,10 +12,75 @@
...
 
+import scala.collection.LinearSeq
 import scala.collection.generic.CanBuildFrom
 import scala.collection.mutable.Builder
 import scala.collection.{immutable => i, mutable => m}
 
+/* builder optimized for a single ++= call, which returns identity on result if possible
+ * and defers to the underlying builder if not.
+ */
+private final class IdentityPreservingSeqBuilder[A](that: Builder[A, Seq[A]])
```
NthPortal: why not the following, so you can use the same class for both?

```scala
private final class IdentityPreservingSeqBuilder[A, CC[X] <: Seq[X]: ClassTag](that: Builder[A, CC[A]])
```

martijnhoekstra: That's a neat trick, I used it in https://github.com/scala/scala-collection-compat/pull/238/commits/ee6b3367c0e83497b0c7e80e0a7f95c8ccfda670

There is a lot of typecasting going on, which I'm not very experienced with, and I'm not sure this is the best way to do it, or the casts are neater elsewhere (e.g. on the inner builders) so please do review for correctness and style.

compat/src/main/scala-2.11_2.12/scala/collection/compat/CompatImpl.scala
```diff
@@ -12,10 +12,47 @@
...
+ * and defers to the underlying builder if not.
+ */
+private final class IdentityPreservingBuilder[A, CC[X] <: TraversableOnce[X]](that: Builder[A, CC[A]])(implicit ct: ClassTag[CC[A]])
+    extends Builder[A, CC[A]] {
+  var collection: CC[A] = null.asInstanceOf[CC[A]]
+  var ruined = false
+
+  final override def ++=(elems: TraversableOnce[A]): this.type =
+      elems match {
+        case ct(ca) if (collection == null && !ruined) =>  {
```
NthPortal: I think you can just do `case ca: CC[A] if (...) =>` and the compiler figures it out

compat/src/main/scala-2.11_2.12/scala/collection/compat/CompatImpl.scala
```diff
@@ -12,10 +12,47 @@
...
+          ruined = true
+          if (collection != null) that ++= collection
+          that ++= elems
+          collection = null.asInstanceOf[CC[A]]
+          this
+        }
+      }
+
+  final def +=(elem: A): this.type = {
+    collection = null.asInstanceOf[CC[A]]
```
NthPortal: does this need to do `if (collection != null) that ++= collection`? should that logic be extracted out into a(n inline?) method?

```scala
private def ruin(): Unit = {
  ruined = true
  val coll = collection
  if (coll != null) {
    that ++= coll
    collection = null.asInstanceOf[CC[A]]
  }
}
```

compat/src/main/scala-2.11_2.12/scala/collection/compat/CompatImpl.scala
```diff
@@ -12,10 +12,47 @@
...
+    ruined = true
+    that += elem
+    this
+  }
+  final def clear(): Unit = {
+    collection = null.asInstanceOf[CC[A]]
+    if (ruined) that.clear()
```
NthPortal: also `ruined = false`?

compat/src/main/scala-2.11_2.12/scala/collection/compat/PackageShared.scala
```diff
@@ -45,8 +45,12 @@ private[compat] trait PackageShared {
   }
 
   implicit def genericCompanionToCBF[A, CC[X] <: GenTraversable[X]](
-      fact: GenericCompanion[CC]): CanBuildFrom[Any, A, CC[A]] =
-    simpleCBF(fact.newBuilder[A])
+      fact: GenericCompanion[CC]): CanBuildFrom[Any, A, CC[A]] = {
+    val builder = if (fact == Seq) new IdentityPreservingBuilder[A, Seq](Seq.newBuilder[A])
+                  else if (fact == LinearSeq) new IdentityPreservingBuilder[A, LinearSeq](LinearSeq.newBuilder[A])
```
NthPortal: do we also need `immutable` variants? `GenSeq`?

would it then be more performant to use an identity hash map?

thoughts @julienrf?

martijnhoekstra: `GenSeq` is gone, right?

compat/src/main/scala-2.11_2.12/scala/collection/compat/CompatImpl.scala
martijnhoekstra: saved!

compat/src/main/scala-2.11_2.12/scala/collection/compat/PackageShared.scala
NthPortal: good point

compat/src/main/scala-2.11_2.12/scala/collection/compat/CompatImpl.scala
julienrf: Unfortunately not. I suggest removing the `ClassTag` parameter but taking instead the partial function that performs the runtime test, as a parameter.

martijnhoekstra: @julienrf are you sure? With this format tests pass (though start failing when I go to Nth's proposal). What's the failure mode?

compat/src/main/scala-2.11_2.12/scala/collection/compat/CompatImpl.scala
martijnhoekstra: I did a general update of ruination, which is, I think, a bit tighter still.

compat/src/main/scala-2.11_2.12/scala/collection/compat/PackageShared.scala
martijnhoekstra: I now do the variants `c.Seq`, `i.Seq`, `c.LinearSeq`, `i.LinearSeq`.

compat/src/main/scala-2.11_2.12/scala/collection/compat/CompatImpl.scala
NthPortal: I guess it just doesn't work with HKT. It works fine with normal kinded types:

```scala
Welcome to Scala 2.12.8 (OpenJDK 64-Bit Server VM, Java 1.8.0_212).

scala> def test[A: reflect.ClassTag](elem: Any): Boolean = elem match {
     |   case a: A => true
     |   case _ => false
     | }
test: [A](elem: Any)(implicit evidence$1: scala.reflect.ClassTag[A])Boolean

scala> test[Int]("foo")
res0: Boolean = false

scala> test[Int](1)
res1: Boolean = true

scala> test[String](1)
res2: Boolean = false

scala> test[String]("foo")
res3: Boolean = true
```

martijnhoekstra:
I can't reconcile the documentation with the observation that the extractor
object works, but the plain type match doesn't.

Maybe there is a bug in the implementation.

compat/src/main/scala-2.11_2.12/scala/collection/compat/CompatImpl.scala
julienrf: Wow interesting, TIL that matching on a universally quantified type `A` does work if there is an implicit `ClassTag` in scope.

However, I suggest doing the following instead, I’m always unsure how reliable are `ClassTag`s on Scala.js.

~~~ scala
class IdentityPreservingBuilder[A, CC[X] <: TraversableOnce[X]](
  that: m.Builder[A, CC[A]],
  extractCC: PartialFunction[TraversableOnce[A], CC[A]]
) extends m.Builder[A, CC[A]] {

  private val extracted = new PartialFunctionAsExtractor(extractCC) // Unfortunately, #7111 is not available on 2.12 and 2.11

...

  override def ++=(elems: TraversableOnce[A]): this.type =
    elems match {
      case extracted(ca) if (collection == null && !ruined) =>
        collection = ca
        this
      case _ =>
        ruin()
        that ++= elems
        this
    }

...

}

private class PartialFunctionAsExtractor[CC[_], A](pf: PartialFunction[TraversableOnce[A], CC[A]]) {
  def unapply(elems: TraversableOnce[A]): Option[CC[A]] = pf.lift(elems)
}
~~~

And then:

~~~ diff
- case c.Seq | i.Seq => new IdentityPreservingBuilder[A, i.Seq](i.Seq.newBuilder[A])
+ case c.Seq | i.Seq =>
+   new IdentityPreservingBuilder[A, i.Seq](i.Seq.newBuilder[A], { case xs: i.Seq[A] => xs })
~~~

compat/src/main/scala-2.11_2.12/scala/collection/compat/CompatImpl.scala
```diff
@@ -12,10 +12,55 @@
...
+  var collection: CC[A] = null.asInstanceOf[CC[A]]
+  var ruined = false
+
+  private[this] def ruin(): Unit = {
+    if(collection != null) that ++= collection
+    collection = null.asInstanceOf[CC[A]]
+    ruined = true
+  }
+
+  final override def ++=(elems: TraversableOnce[A]): this.type =
```
julienrf: `final` is redundant here because the class itself is already `final`

compat/src/main/scala-2.11_2.12/scala/collection/compat/CompatImpl.scala
```diff
@@ -12,10 +12,55 @@
...
+        this
+      }
+      case _ => {
+        ruin()
+        that ++= elems
+        this
+      }
+    }
+
+  final def +=(elem: A): this.type = {
```
julienrf: Same remark here about `final`.

compat/src/main/scala-2.11_2.12/scala/collection/compat/CompatImpl.scala
```diff
@@ -12,10 +12,55 @@
...
+    ruin()
+    that += elem
+    this
+  }
+
+  final def clear(): Unit = {
```
julienrf: Same here (`final`).

compat/src/main/scala-2.11_2.12/scala/collection/compat/CompatImpl.scala
```diff
@@ -12,10 +12,55 @@
...
+    collection = null.asInstanceOf[CC[A]]
+    if (ruined) that.clear()
+    ruined = false
+  }
+
+  final def result(): CC[A] = if(collection == null) that.result() else collection
```
julienrf: Same here (`final`).

compat/src/main/scala-2.11_2.12/scala/collection/compat/CompatImpl.scala
```diff
@@ -12,10 +12,55 @@
...
+
+  private[this] def ruin(): Unit = {
+    if(collection != null) that ++= collection
+    collection = null.asInstanceOf[CC[A]]
+    ruined = true
+  }
+
+  final override def ++=(elems: TraversableOnce[A]): this.type =
+    elems match {
+      case ct(ca) if (collection == null && !ruined) =>  {
```
julienrf: Braces are not needed on the right hand side of a pattern.

compat/src/main/scala-2.11_2.12/scala/collection/compat/CompatImpl.scala
```diff
@@ -12,10 +12,55 @@
...
 
+import scala.reflect.ClassTag
 import scala.collection.generic.CanBuildFrom
 import scala.collection.mutable.Builder
 import scala.collection.{immutable => i, mutable => m}
 
+/* builder optimized for a single ++= call, which returns identity on result if possible
+ * and defers to the underlying builder if not.
+ */
+private final class IdentityPreservingBuilder[A, CC[X] <: TraversableOnce[X]](that: Builder[A, CC[A]])(implicit ct: ClassTag[CC[A]])
```
julienrf: Do you mind writing `m.Builder` everywhere instead of `Builder`?

compat/src/main/scala-2.11_2.12/scala/collection/compat/CompatImpl.scala
martijnhoekstra: isn't that pretty much exactly what `ClassTag[i.Seq[_]]#unapply.unlift` is?

@sjrd could you shine your light on classtag reliability?

compat/src/main/scala-2.11_2.12/scala/collection/compat/CompatImpl.scala
martijnhoekstra: done

compat/src/main/scala-2.11_2.12/scala/collection/compat/CompatImpl.scala
julienrf: > isn't that pretty much exactly what `ClassTag[i.Seq[_]]#unapply.unlift` is?

Yes, that’s equivalent.

sjrd: The non-deprecated parts of `ClassTag`s work in Scala.js exactly as on the JVM.

The deprecated method `<:<`, which really is a method of the `Manifest` API, doesn't link on Scala.js.

PR #252, opened 2019-09-13
New scala-module-plugin with sbt-ci-release / travisci / dynver / header

lrytz:


build.sbt
```diff
@@ -52,16 +50,6 @@ lazy val scala211 = "2.11.12"
 lazy val scala212 = "2.12.8"
 lazy val scala213 = "2.13.0"
 
-scalaVersionsByJvm in ThisBuild := {
```
lrytz: maybe this was still in use to ensure we don't publish on jdk 11, so we'll have to add a condition to the `ci` command.

lrytz:
Merging, I'll push a test tag to see how it does.

lrytz:
What's missing, compared to other modules, is the ability to re-release an existing tag against a new binary incompatible scala version (https://github.com/scala/scala-xml/blob/master/build.sh#L13-L17). (It was not removed by this PR, the functionality was not there).

lrytz:
Needs a bit more tweaking, my test tag didn't failed to release to sonatype https://travis-ci.org/scala/scala-collection-compat/builds/585619869.

i'll experiment in my fork.

PR #268, opened 2019-11-22
Build for Scala.js 1.0.0-RC1

xerial:


julienrf:
It looks like we should disable mima when building for Scala.js 1.0.0-RC1

xerial:
I don't know why, but Travis CI environment is broken.

lrytz:
cleared the travis caches and re-started the builds (cause: same as https://github.com/scala/scala/pull/8528), the results should be acurate now.

build.sbt
```diff
@@ -14,7 +14,13 @@ lazy val commonSettings = Seq(
                                                  |See the NOTICE file distributed with this work for
                                                  |additional information regarding copyright ownership.
                                                  |""".stripMargin)),
-  scalaModuleMimaPreviousVersion := Some("2.1.1"),
+  scalaModuleMimaPreviousVersion := {
+    // We need to create a release version first to use MiMa
+    if(sys.env.get("SCALAJS_VERSION").exists(_.startsWith("1.0.0-M")))
```
julienrf: Should be `1.0.0-`, no? (without the `M`)

ekrich: I was thinking that as well if we are changing to `RC1` ...

julienrf: Also, it should be `forall` rather than `exists` because we still want to enable MiMa when building JVM artifacts.

xerial:
Applied the suggested fixes. At least, compatXXXJs/tests worked in my machine.

build.sbt
julienrf: Uh, I’m sorry but my reasoning was wrong, using `exists` as you initially did was the correct thing to do…

julienrf:
Could you please sign the Scala CLA?
https://www.lightbend.com/contribute/cla/scala

xerial:
I think I have already signed to that, but let me double check.

xerial:
Signing is done.

julienrf:
Thanks @xerial! It seems that something is wrong with the CLA system… Does anyone have an idea of what we can do? (@lrytz or @SethTisue?)

lrytz:
[`checkCLA.sh`](https://github.com/scala/scala-collection-compat/blob/master/admin/checkCLA.sh#L7) does

```
curl -u dummy4dummy:dummy2dummy -s "https://api.github.com/repos/scala/scala-collection-compat/pulls/268"
```

which gives me, on my machine

```
{
  "message": "Bad credentials",
  "documentation_url": "https://developer.github.com/v3"
}
```

If I remove the dummy login, it works

```
$> curl -s "https://api.github.com/repos/scala/scala-collection-compat/pulls/268"
{
  "url": "https://api.github.com/repos/scala/scala-collection-compat/pulls/268",
  "id": 344705263,
  "node_id": "MDExOlB1bGxSZXF1ZXN0MzQ0NzA1MjYz",
  "html_url": "https://github.com/scala/scala-collection-compat/pull/268",
  "diff_url": "https://github.com/scala/scala-collection-compat/pull/268.diff",
  "patch_url": "https://github.com/scala/scala-collection-compat/pull/268.patch",
  "issue_url": "https://api.github.com/repos/scala/scala-collection-compat/issues/268",
...
```

I saw the same in some other repo while ago, don't remember which.. removed the login.

lrytz:
hmm, scratch that, i had an old checkout locally; i did remove the dummy login in this repo https://github.com/scala/scala-collection-compat/commit/78ce8686de4fe3b8213cba98ad896d895b31d4b2.

lrytz:
did a few re-runs, it's green now. not sure what's behind the cla check failure..

julienrf:
Thank you all for the team work ;)

PR #273, opened 2019-12-03
Import travis caching config and JDK install from scala-dev

lrytz:


lrytz:
Here's the reason why the CLA check fails regularly:

https://travis-ci.org/scala/scala-collection-compat/jobs/620201244?utm_medium=notification&utm_source=github_status

```
{ "message": "API rate limit exceeded for 35.202.245.105. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)", "documentation_url": "https://developer.github.com/v3/#rate-limiting" }
```

lrytz:
Anyone has an idea how to work around this? We could create an access token for some user, but we cannot store that in an encrypted variable on travis, as those are not available to pull request builds. Maybe a small proxy app that runs somewhere and does the check?

lrytz:
https://colineberhardt.github.io/cla-bot/ seems to cover everything we need, i'll give it a try. it can be enabled all of the repos in the scala organization.

lrytz:
almost works for us, but would need a few tweaks, so we'd have to propose some changes or run our own. anyway - suggestions welcome, looking for a simple solution :)

lrytz:
Merging this for now, will come back to the CLA checker issue.

SethTisue:
on recent PRs and publishing jobs I've noticed occasional CLA check fails, but not a high percentage. they go away by hitting the restart button in Travis-CI. it's a bit annoying, but not a serious hindrance

lrytz:
Created https://github.com/scala/scala-collection-compat/issues/280 for the CLA discussion

Issue #276, opened 2019-12-11
release 2.1.3


SethTisue:
for two reasons:
* we should (as @ekrich has pointed out) publish for Scala.js 1.0.0-RC1
* to address #245 (scala-collection-migrations didn't publish successfully when we did 2.1.2)

this will be the first time we release this module with sbt-ci-release instead of the old module releasing process, so I'm half expecting it will go wrong. and there are some ways in which it might go wrong that would still involve scala-steward jumping on it before you can say "holy runaway automation, Batman"

so I'd like to do a 2.1.3-RC1 first, see if everything publishes, and have Eric verify that the artifacts are usable with Scala.js 1.0.0-RC1. if that all goes well, I'll follow up with a proper 2.1.3 release

(2.1.2 release ticket was #214)

SethTisue:
> this will be the first time we release this module with sbt-ci-release instead of the old module releasing process, so I'm half expecting it will go wrong

though I see now that @lrytz already did a v2.1.3-TEST at https://travis-ci.org/scala/scala-collection-compat/builds/585968641 (history at #252, #255). that's encouraging. regardless, I would like to do an RC

SethTisue:
* tag: https://github.com/scala/scala-collection-compat/releases/tag/v2.1.3-RC1
* release build: https://travis-ci.org/scala/scala-collection-compat/builds/623505472

gah `[error] java.io.FileNotFoundException: /home/travis/.sbt/gpg/secring.asc (No such file or directory)`

SethTisue:
it's about my bedtime. @lrytz perhaps you could take a look if you're not too busy, since you've been over this before. otherwise I'll come back to it myself (I am no longer underwater, I have a decent amount of time this week and next)

lrytz:
Ah, also in the travis log:

The good version has:

```
Running CI: 
compat211/clean
compat211/test
compat211/publishLocal
set pgpSecretRing := pgpSecretRing.value
set pgpPublicRing := pgpPublicRing.value
compat211/publishSigned
sonatypePrepare
sonatypeBundleUpload
sonatypeClose
```


The bad one:

```
Running CI: 
compat211/clean
compat211/test
compat211/publishLocal
compat211/publishSigned
sonatypePrepare
sonatypeBundleUpload
sonatypeClose
```

So it's indeed the workaround that's missing.

lrytz:
[...]

PR #282, opened 2019-12-12
update readme and build for 2.1.3 release

SethTisue:
(this won't go green until all of the 2.1.3 artifacts have hit Maven Central)


SethTisue:
LOL, I spoke too soon, `MiMa will NOT run because the previous artifact "org.scala-lang.modules" % "scala-collection-compat_2.12" % "2.1.3" could not be resolved (note the binary Scala version)` but the jobs keep on trucking

regardless, we shouldn't merge this until #276 is completely closed out

PR #287, opened 2020-01-03
Add filterKeys extension method to MapView

lavrov:


lavrov:
I forgot to sign CLA. Just did it.

PR #288, opened 2020-01-10
Add groupMap and groupMapReduce extensions

### LINKED ISSUE #11: Add groupMap and groupMapReduce
So that users can migrate the following pattern:

~~~ scala
iterable.groupBy(f).mapValues(g).toMap
~~~

Into just:

~~~ scala
iterable.groupMap(f)(g)
~~~

NthPortal:
Fixes #11

### DIFF
--- a/compat/src/main/scala-2.11_2.12/scala/collection/compat/PackageShared.scala
+++ b/compat/src/main/scala-2.11_2.12/scala/collection/compat/PackageShared.scala
@@ -243,6 +243,36 @@ class TraversableExtensionMethods[A](private val self: c.Traversable[A]) extends
   def iterableFactory: GenericCompanion[Traversable] = self.companion
 }
 
+class TraversableLikeExtensionMethods[A, Repr](private val self: c.GenTraversableLike[A, Repr])
+    extends AnyVal {
+
+  def groupMap[K, B, That](key: A => K)(f: A => B)(
+      implicit bf: CanBuildFrom[Repr, B, That]): Map[K, That] = {
+    val map = m.Map.empty[K, m.Builder[B, That]]
+    for (elem <- self) {
+      val k    = key(elem)
+      val bldr = map.getOrElseUpdate(k, bf(self.repr))
+      bldr += f(elem)
+    }
+    val res = Map.newBuilder[K, That]
+    for ((k, bldr) <- map) res += ((k, bldr.result()))
+    res.result()
+  }
+
+  def groupMapReduce[K, B](key: A => K)(f: A => B)(reduce: (B, B) => B): Map[K, B] = {
+    val map = m.Map.empty[K, B]
+    for (elem <- self) {
+      val k = key(elem)
+      val v = map.get(k) match {
+        case Some(b) => reduce(b, f(elem))
+        case None => f(elem)
+      }
+      map.put(k, v)
+    }
+    map.toMap
+  }
+}
+
 class MapViewExtensionMethods[K, V, C <: scala.collection.Map[K, V]](
     private val self: IterableView[(K, V), C])
     extends AnyVal {
--- a/compat/src/main/scala-2.11/scala/collection/compat/package.scala
+++ b/compat/src/main/scala-2.11/scala/collection/compat/package.scala
@@ -12,4 +12,11 @@
 
 package scala.collection
 
-package object compat extends compat.PackageShared
+import scala.collection.generic.IsTraversableLike
+
+package object compat extends compat.PackageShared {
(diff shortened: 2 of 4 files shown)


NthPortal:
I think there's some MiMa twiddling that needs to happen, but I believe this code should pretty much work

compat/src/main/scala-2.11_2.12/scala/collection/compat/PackageShared.scala
```diff
@@ -243,6 +250,40 @@ class TraversableExtensionMethods[A](private val self: c.Traversable[A]) extends
   def iterableFactory: GenericCompanion[Traversable] = self.companion
 }
 
+class TraversableLikeExtensionMethods[A, Repr](private val self: c.GenTraversableLike[A, Repr])
+    extends AnyVal {
+
+  def groupMap[K, B, VThat, That](key: A => K)(f: A => B)(
+      implicit traversable: IsTraversableLikeAux[A, Repr],
+      innerBf: CanBuildFrom[Repr, B, VThat],
+      outerBf: CanBuildFrom[c.GenMap[K, Repr], (K, VThat), That]): That = {
```
julienrf: I think we should fix the return type to `immutable.Map[K, VThat]`, like we do in the scala-library.

compat/src/main/scala-2.11_2.12/scala/collection/compat/PackageShared.scala
```diff
@@ -243,6 +250,40 @@ class TraversableExtensionMethods[A](private val self: c.Traversable[A]) extends
...
+      val innerB = innerBf(repr)
+      for (a <- traversable.conversion(repr)) innerB += f(a)
+      outerB.+=((k, innerB.result()))
+    }
+    outerB.result()
+  }
+
+  def groupMapReduce[K, B, That](key: A => K)(f: A => B)(reduce: (B, B) => B)(
+      implicit traversable: IsTraversableLikeAux[A, Repr],
+      bf: CanBuildFrom[c.GenMap[K, Repr], (K, B), That]): That = {
```
julienrf: Here too, we could simply fix the return type to `immutable.Map[K, B]`.

compat/src/main/scala-2.11_2.12/scala/collection/compat/PackageShared.scala
NthPortal: I don't have a problem with that; however, it's not what `groupBy` does

julienrf: I was referring to [these implementations](https://github.com/scala/scala/blob/v2.13.1/src/library/scala/collection/Iterable.scala#L538-L630), I think that using as few `CanBuildFrom` as possible will make the code more readable :)

NthPortal: I rewrote it to just use the 2.13 implementation, which massively simplified things

compat/src/main/scala-2.11_2.12/scala/collection/compat/PackageShared.scala
```diff
@@ -30,6 +30,8 @@ private[compat] trait PackageShared {
    */
   type Factory[-A, +C] = CanBuildFrom[Nothing, A, C]
 
+  type IsTraversableLikeAux[A1, Repr] = IsTraversableLike[Repr] { type A = A1 }
```
julienrf: You don’t seem to be using this type alias anymore.

julienrf:
Unfortunately, this PR is not backward compatible (see https://stackoverflow.com/a/18369508/561721 for an explanation of why).

I think we could create a `2.x` branch where we could apply backward-compatible changes (if needed).

And branch `master` could be the place where we prepare the 3.x release…

What do others think?

lrytz:
Maybe the `toTraversableLikeExtensionMethods` method can be duplicated and added directly to the `compat` objects, instead of the `PackageShared` super trait? That would avoid the binary incompatibility on 2.11.

NthPortal:
for future versions, can `PackageShared` be an `abstract class` rather than a `trait` (thus sidestepping the problem)?

NthPortal:
@lrytz yup, I'll do that

julienrf:
@NthPortal Yes, that should work (see #292)

compat/src/main/scala-2.11_2.12/scala/collection/compat/PackageShared.scala
NthPortal: good catch!

NthPortal:
cleaned up, squashed, and ready for merge

PR #290, opened 2020-01-11
Backport LazyList

NthPortal:
Still TODO:

- [x] Fix compilation on 2.11
- [x] ~~Fix serialization~~

### DIFF
--- a/compat/src/main/scala-2.11_2.12/scala/collection/compat/immutable/LazyList.scala
+++ b/compat/src/main/scala-2.11_2.12/scala/collection/compat/immutable/LazyList.scala
@@ -1,0 +1,1511 @@
+/*
+ * Scala (https://www.scala-lang.org)
+ *
+ * Copyright EPFL and Lightbend, Inc.
+ *
+ * Licensed under Apache License 2.0
+ * (http://www.apache.org/licenses/LICENSE-2.0).
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+package scala.collection.compat.immutable
+
+import java.io.{ObjectInputStream, ObjectOutputStream}
+
+import scala.annotation.tailrec
+import scala.annotation.unchecked.{uncheckedVariance => uV}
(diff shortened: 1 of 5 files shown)


NthPortal:
I wish I could mark this as WIP again, cuz boy is it not done.
help welcome

julienrf:
@NthPortal Could you explain what remains to be done?

NthPortal:
@julienrf port GC tests, and make all the tests pass in all versions (or comment out/fix those that can't due to compiler bugs that weren't fixed until 2.13)

NthPortal:
if someone can take a look at the last two little bits and make them compile, that would be hugely appreciated

NthPortal:
I've given up on getting the serialization test to work across versions. it will just have to be specific to certain versions (2.11 and 2.13 JVM)

NthPortal:
it seems GC tests don't work on JS (idk why I missed this before). Is there a good way to let them run only on JVM? or can they be tweaked to use something other than `Thread.sleep`?

NthPortal:
The docs for `System.gc()` say that when the method returns, the GC will already have been run (to the extent the "best effort" could), so it's not clear to me that the `Thread.sleep(10)` is necessary at all.

However, even without the `Thread.sleep`, the tests for scala.js still fail (probably because `System.gc()` is a NOP in scala.js). While JS engines *do* tend to have garbage collectors, there doesn't appear to be a way to manually invoke them (which is fair, from a browser design perspective).

yannbolliger:
The fix for this is here: https://github.com/NthPortal/scala-collection-compat/pull/1

julienrf:
Thank you both!

nilskp:
Do we know when this will be released?

julienrf:
Hey, we should indeed cut a release soon, thanks for asking :smiley: 
I think it would be good to have Lukas’ work in it. Let’s wait for a couple of weeks to see if we manage to merge it.

lrytz:
Is there a reason not to release often? It's little work, we have everything automated, no?

julienrf:
I think last time we had an issue and had to manually publish things. But I agree that in principle we could release more often.

PR #292, opened 2020-01-13
Prepare for working on version 3.x

julienrf:
- Disable MiMa checks
- Turn `trait PackageShared` into a `class`, so that we can later add methods without breaking binary compatibility.


julienrf:
Let’s not merge this PR unless we have to. We are currently exploring workarounds to stay binary compatible.

NthPortal:
> Let’s not merge this PR unless we have to. We are currently exploring workarounds to stay binary compatible.

see https://github.com/scala/scala-collection-compat/pull/288#issuecomment-573658463

PR #295, opened 2020-01-18
Allow chaining of filterKeys and mapValues

### LINKED ISSUE #294: IterableView#filterKeys and #mapValues cannot be chained
With the current extension methods, (as of 2badce8debcc6d4aba1b9113810315cc032192a6), you cannot chain `filterKeys` and `mapValues` on an `IterableView[_, _ <: Map]`. I believe the reason for this is that both take a `CanBuildFrom` and return a `That`, unlike their corresponding 2.12 methods. The `filterKeys` extension is even particularly egregious, as `filter` methods always return the same type and do not ever take a `CanBuildFrom`.

The methods are currently defined as:

```scala
class MapViewExtensionMethods[K, V, C <: scala.collection.Map[K, V]](
    private val self: IterableView[(K, V), C])
    extends AnyVal {

  def mapValues[W, That](f: V => W)(
      implicit bf: CanBuildFrom[IterableView[(K, V), C], (K, W), That]): That =
    self.map[(K, W), That] { case (k, v) => (k, f(v)) }

  def filterKeys[That](p: K => Boolean)(
      implicit bf: CanBuildFrom[IterableView[(K, V), C], (K, V), That]): That =
    self.collect[(K, V), That] { case (k, v) if p(k) => (k, v) }
}
```

The following definitions do not use `CanBuildFrom` and can chain with each other:

```scala
class MapViewExtensionMethods[K, V, C <: scala.collection.Map[K, V]](
    private val self: IterableView[(K, V), C])
    extends AnyVal {

  def mapValues[W](f: V => W): IterableView[(K, W), C] =
    // the implementation of `self.map` also casts the result
    self.map({ case (k, v) => (k, f(v)) }).asInstanceOf[IterableView[(K, W), C]]

  def filterKeys(p: K => Boolean): IterableView[(K, V), C] =
    self.filter { case (k, _) => p(k) }
}
```

Unfortunately, `mapValues` has already been in previous releases, so its signature cannot be changed without a new release. However, we can still change the implementation of `filterKeys` (at least as of the aforementioned commit), as it has not been in any release yet. This allows the (I believe more common) use of `.view.filterKeys(...).mapValues(...)`, though not `.view.mapValues(...).filterKeys(...)`.

NthPortal:
Fix implementation of `IterableView#filterKeys` to not require an
implicit CBF, and add a commented-out implementation of
`IterableView#mapValues` that does not require an implicit CBF.

Add an implicit method to create a CBF so that
`IterableView#mapValues` can be chained in front of
`IterableView#filterKeys`.

Fixes #294

### DIFF
--- a/compat/src/main/scala-2.11_2.12/scala/collection/compat/PackageShared.scala
+++ b/compat/src/main/scala-2.11_2.12/scala/collection/compat/PackageShared.scala
@@ -354,7 +354,15 @@ class MapViewExtensionMethods[K, V, C <: scala.collection.Map[K, V]](
       implicit bf: CanBuildFrom[IterableView[(K, V), C], (K, W), That]): That =
     self.map[(K, W), That] { case (k, v) => (k, f(v)) }
 
-  def filterKeys[That](p: K => Boolean)(
-      implicit bf: CanBuildFrom[IterableView[(K, V), C], (K, V), That]): That =
-    self.collect[(K, V), That] { case (k, v) if p(k) => (k, v) }
+  // TODO: Replace the current implementation of `mapValues` with this
+  //       after major version bump when bincompat can be broken.
+  //       At the same time, remove `canBuildFromIterableViewMapLike`
+  /*
+  def mapValues[W](f: V => W): IterableView[(K, W), C] =
+    // the implementation of `self.map` also casts the result
+    self.map({ case (k, v) => (k, f(v)) }).asInstanceOf[IterableView[(K, W), C]]
+   */
+
+  def filterKeys(p: K => Boolean): IterableView[(K, V), C] =
+    self.filter { case (k, _) => p(k) }
 }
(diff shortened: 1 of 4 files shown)


compat/src/main/scala-2.11_2.12/scala/collection/compat/PackageShared.scala
```diff
@@ -284,7 +284,14 @@ class MapViewExtensionMethods[K, V, C <: scala.collection.Map[K, V]](
...
+  // TODO: replace the current implementation of `mapValues` with this
+  //       after major version bump when bincompat can be broken
+  /*
+  def mapValues[W](f: V => W): IterableView[(K, W), C] =
+    // the implementation of `self.map` also casts the result
+    self.map({ case (k, v) => (k, f(v)) }).asInstanceOf[IterableView[(K, W), C]]
+   */
+
+  def filterKeys(p: K => Boolean): IterableView[(K, V), C] =
+    self.filter { case (k, _) => p(k) }
```
julienrf: I don’t understand how this change can be backward compatible… There is no `CanBuildFrom` parameter anymore.

NthPortal: I don't believe `filterKeys` has been released yet, so as far as artifacts on maven are concerned, it doesn't exist

julienrf: Ah, yes that explains why :)

julienrf:
Thanks @NthPortal!

PR #318, opened 2020-04-15
Update sbt-scala-module to 2.1.4

lrytz:
This new version includes the travis job number in the sonatype repo
description, which ensures that travis jobs don't interfere by writing
to the same repo.

In particular, for 2.12.11, there are two travis jobs in this repo:
one for the migration rules, one for the compat library.

This is why the 2.12 release was missing last time:
https://contributors.scala-lang.org/t/this-is-not-the-release-announcement-of-scala-js-1-0-0/4021/9


lrytz:
I tested this with a new tag in my travis-test project (https://github.com/lrytz/travis-test/releases/tag/v0.3.0-RC1), here's how it looks on sonatype:

image

PR #319, opened 2020-04-15
Backport Using and new ControlThrowable

NthPortal:
Backport `scala.util.Using`.

Backport `scala.util.control.ControlThrowable` from
2.13 (which does not suppress exceptions or have a
stack trace) as `scala.util.control.compat.ControlThrowable`
(it still extends previous Scala versions'
`scala.util.control.ControlThrowable`).

### DIFF
--- a/compat/src/main/scala-2.11_2.12/scala/util/Using.scala
+++ b/compat/src/main/scala-2.11_2.12/scala/util/Using.scala
@@ -1,0 +1,403 @@
+/*
+ * Scala (https://www.scala-lang.org)
+ *
+ * Copyright EPFL and Lightbend, Inc.
+ *
+ * Licensed under Apache License 2.0
+ * (http://www.apache.org/licenses/LICENSE-2.0).
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+package scala.util
+
+import scala.util.control.{ControlThrowable, NonFatal}
+
+/** A utility for performing automatic resource management. It can be used to perform an
+ * operation using resources, after which it releases the resources in reverse order
+ * of their creation.
+ *
+ * ==Usage==
+ *
+ * There are multiple ways to automatically manage resources with `Using`. If you only need
+ * to manage a single resource, the [[Using.apply `apply`]] method is easiest; it wraps the
+ * resource opening, operation, and resource releasing in a `Try`.
+ *
+ * Example:
+ * {{{
+ * import java.io.{BufferedReader, FileReader}
+ * import scala.util.{Try, Using}
+ *
+ * val lines: Try[Seq[String]] =
+ *   Using(new BufferedReader(new FileReader("file.txt"))) { reader =>
+ *     Iterator.continually(reader.readLine()).takeWhile(_ != null).toSeq
+ *   }
+ * }}}
+ *
+ * If you need to manage multiple resources, [[Using.Manager$.apply `Using.Manager`]] should
+ * be used. It allows the managing of arbitrarily many resources, whose creation, use, and
+ * release are all wrapped in a `Try`.
+ *
+ * Example:
+ * {{{
+ * import java.io.{BufferedReader, FileReader}
+ * import scala.util.{Try, Using}
+ *
+ * val lines: Try[Seq[String]] = Using.Manager { use =>
+ *   val r1 = use(new BufferedReader(new FileReader("file1.txt")))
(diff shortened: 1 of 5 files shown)


compat/src/main/scala-2.11_2.12/scala/util/Using.scala
```diff
@@ -0,0 +1,401 @@
...
+      body(resource)
+    } catch {
+      case t: Throwable =>
+        toThrow = t
+        null.asInstanceOf[A] // compiler doesn't know `finally` will throw
+    } finally {
+      if (toThrow eq null) releasable.release(resource)
+      else {
+        try releasable.release(resource)
+        catch { case other: Throwable => toThrow = preferentiallySuppress(toThrow, other) } finally throw toThrow
```
NthPortal, line 278: why is this `finally` not on its own line? formatting is hard

SethTisue: ha, is scalafmt doing that? that seems.... suboptimal

NthPortal: presumably

I'll see if I can faff around with it and get it to not look horrible like that

compat/src/main/scala-2.11_2.12/scala/util/control/compat/ControlThrowable.scala
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * Copyright EPFL and Lightbend, Inc.
+ *
+ * Licensed under Apache License 2.0
+ * (http://www.apache.org/licenses/LICENSE-2.0).
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+package scala.util.control.compat
```
julienrf, line 12: Why do we need to put the definition of `ControlThrowable` in this `scala.util.control.compat` package as opposed to just `scala.util.control`?

NthPortal: because `scala.util.control` already contains `ControlThrowable`. also it seemed consistent to me with what was done with `scala.collection.compat`

julienrf: OK I see. Sorry, I haven’t been using `ControlThrowable` much so it’s not clear to me why we want to have this `scala.util.control.compat.ControlThrowable` subtype of `scala.util.control.ControlThrowable`. Do you mind elaborating? Are end-users going to use it, or is it only used internally by `Using`?

NthPortal: the issue is, in order to disable suppression behavior in 2.12-, you need to do

```scala
class CustomException extends Throwable("some message", null, false, false) with ControlThrowable
```

This gives you the desired behavior, while also giving you the special treatment you get from things like `NonFatal`.

The problem is, in 2.13, it's an abstract class not a trait, so you can't mix it in. It makes it difficult to get the correct behaviour in a cross-compiling way.

This implementation sidesteps that by being a concrete implementation (an abstract class) in 2.12-, but a type alias for an abstract class in 2.13. It extends `scala.util.control.ControlThrowable` so that it still doesn't get matched by `NonFatal`.

It is not actually used by `Using`. I initially added it so that `UsingTest` could use it (because I was having the issue described above about trait mixin vs abstract class). I ended up not needing to use it, since I decided to special-case `ControlThrowable` in `Using` ([here](https://github.com/scala/scala-collection-compat/pull/319/files#diff-bd53a44e5ba7fa031f39ccd396b35b4bR244)) and never suppress exceptions for it, but I had already written `compat.ControlThrowable` at that point.

julienrf: So, the goal is to have it available for end-users, right? That looks fine to me, although I believe this will need to thoroughly documented.

NthPortal: > So, the goal is to have it available for end-users, right?

correct

> I believe this will need to thoroughly documented.

in what way do you mean? more than just scaladocs?

julienrf: Actually, we don’t really have rich docs… Maybe we should add more documentation, saying what is supported, what is not, what you need to import, etc., but that’s another issue I believe…

PR #320, opened 2020-04-15
Backport scala.util.chaining

NthPortal:


compat/src/test/scala/scala/util/ChainingOpsTest.scala
```diff
@@ -0,0 +1,37 @@
...
+    val result = (1 - 2 - 3)
+      .pipe(times6)
+      .pipe(scala.math.abs)
+
+    assertEquals(24, result)
+  }
+
+//  @Test(expected = classOf[ToolBoxError])
+//  def testNoSelf(): Unit =
+//    runner.run("import scala.util.chaining._; Nil.self")
```
NthPortal, line 36: not sure if there's a good way to keep this test

julienrf: We need to depend on `scala.tools.testkit.RunTesting`? Is this thing published?

compat/src/main/scala-2.11_2.12/scala/util/package.scala
```diff
@@ -0,0 +1,21 @@
...
+ * Licensed under Apache License 2.0
+ * (http://www.apache.org/licenses/LICENSE-2.0).
+ *
+ * See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership.
+ */
+
+package scala
+
+package object util {
```
julienrf, line 15: Does it have to be a `package object`? Could we put the `object chaining` definition in the package `scala.util` instead?

NthPortal: that's an interesting question. I just copied from the stdlib. what are the practical differences between the two?

compat/src/test/scala/scala/util/ChainingOpsTest.scala
NthPortal: it is, apparently (https://repo1.maven.org/maven2/org/scala-lang/scala-testkit/2.13.1/). but it's only published for 2.13.1, which rather defeats the purpose :(

lrytz: It's also considered internal, the API could break at any time.

compat/src/main/scala-2.11_2.12/scala/util/package.scala
lrytz: putting class and object definitions into package objects is an anti-pattern (mostly because bugs are likely), that's an oversight in the std lib

NthPortal: ok, I'll move it out of the package object.

is it binary compatible if we do that in the stdlib as well?

Issue #317, opened 2020-04-15
rename project now that non-collection stuff is allowed


SethTisue:
see discussion on https://github.com/scala/scala-collection-compat/issues/219

name suggestions so far: `scala-2.13-compat`, `scala-stdlib-compat`, `scala-compat`

SethTisue:
`scala-crossbuilding`, maybe

martijnhoekstra:
my bikeshed would be `scala-crosscompat` or mauve.

ekrich:
I like `scala-stdlib-compat` FWIW.

lrytz:
`scala-library-[compat|crossbuilding|crosscompat]` as it's for `scala-library`?

NthPortal:
I'm for `scala-library-compat` or `scala-library-crosscompat`

lrytz:
Seeing PRs trickle in with backports from the 2.13 library (thanks, @NthPortal), I find it a little ironic that the 2.11 / 2.12 standard libraries can acquire new classes (through this module), while the one in 2.13 (and 3.0) can't.

Should we consider extending the scope of this module to `scala-library-future` where we can also add new classes for 2.13 / 3.0?

julienrf:
> Should we consider extending the scope of this module to `scala-library-future` where we can also add new classes for 2.13 / 3.0?

Given that there will be a lot of time before we can get a 2.14/3.1, yes, that would probably be useful.

SethTisue:
I think `scala-library-compat` is good

> Should we consider extending the scope of this module

Sure, I don't see why not.

lrytz:
> Sure, I don't see why not.

We'd have to make it clear that everything added to the 2.13 release here would eventually end up in the standard library, so we'd have to be cust as conservative.

SethTisue:
I have renamed the repo and submitted a pair of PRs: #327 to do the renaming, and #328 to update the readme. once #327 has been merged, I can do the publishing.

SethTisue:
> Should we consider extending the scope of this module to scala-library-future where we can also add new classes for 2.13 / 3.0?

I've made a separate ticket on this: https://github.com/scala/scala-library-compat/issues/329

PR #327, opened 2020-04-30
rename to scala-library-compat

SethTisue:
fixes #317

note that I already renamed the GitHub repo

#328 has the associated README.md changes separately, to be
merged once we've actually published under the new name

this PR disables MiMa, but I left a breadcrumb in #328 to remember to re-enable it later

PR #328, opened 2020-04-30
update readme for library name change

SethTisue:
references #317

not for merge until we've actually published under the new name

notes to self:
* update the Scala community build, too
* re-enable MiMa
* update the Scala 2.13.0 release notes
* tweet about it from scala_lang (or even do a blog post?)


lrytz:
> until we've actually published under the new name

Let's discuss this first: https://github.com/scala/scala-library-compat/issues/331

julienrf:
According to #331 should we revert #327? In all cases, if we agree on not breaking binary compatibility we should re-enable MiMa and do the check against `scala-library-compat:2.1.6`.

SethTisue:
Lukas is pondering how to proceed here.

NthPortal:
it seems to me like this is PR is obsolete given the conclusion to #331. thanks though Seth 💜

PR #353, opened 2020-07-21
Bump versions mentioned in readme instructions

Daenyth:
The versions in the readme don't work for scala 2.12.12, a newer version of scalafix is needed. Following them as written gives
```
[error] (update) sbt.librarymanagement.ResolveException: Error downloading org.scalameta:semanticdb-scalac_2.12.12:4.2.3
```

whereas with the changes, it works


SethTisue:
thank you!

PR #341, opened 2020-06-05
`scala.collection.Map.foreachEntry` backport for 2.11/2.12

ijuma:
Fixes #335.

PR #356, opened 2020-09-03
sigh, rename back to scala-collection-compat (and re-enable MiMa)

SethTisue:
reverts #327

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,7 +14,7 @@ Several levels of contribution are possible!
 
 ### Report a missing case
 
-Create an issue [scala-library-compat/issues](https://github.com/scala/scala-library-compat/issues).
+Create an issue [scala-collection-compat/issues](https://github.com/scala/scala-collection-compat/issues).
 Embrace `diff`s to describe differences between the standard collections and
 the new collection:
 
--- a/NOTICE
+++ b/NOTICE
@@ -1,4 +1,4 @@
-scala-library-compat
+scala-collection-compat
 Copyright (c) 2002-2020 EPFL
 Copyright (c) 2011-2020 Lightbend, Inc.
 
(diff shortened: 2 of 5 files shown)


SethTisue:
MiMa is failing, but only on 2.11. I'm out of work time for today — I'll look into it tomorrow if nobody beats me to it.

SethTisue:
The failure is due to the addition of `toMapExtensionMethods` in #341. #205 explains why it only fails on 2.11 (trait encoding difference). 

I believe we can filter this out because the trait in question, `PackageShared`, is `private[compat]`.

SethTisue:
#362 has MiMa followups

NthPortal:
@SethTisue I think that method should be moved directly into the package object (which duplicates it because there are separate files for 2.11 and 2.12, but there's not much to do about that). that's what we've been doing for other things

SethTisue:
@NthPortal oops, I was going to look at that before release, but it slipped my mind. is it too late now, because bincompat?

NthPortal:
I... don't know enough about bincompat to say what we should do

SethTisue:
> I... don't know enough about bincompat to say what we should do

Before I think too hard about that, I'd like to understand why you are suggesting the method should be moved directly into the package object.

> which duplicates it because there are separate files for 2.11 and 2.12, but there's not much to do about that

I'm confused — isn't avoiding such duplication exactly what `PackageShared` is for?

NthPortal:
the problem is, adding methods to `PackageShared` isn't binary compatible, I believe. cc @julienrf

SethTisue:
ah, thanks for surfacing that. @julienrf @lrytz was I wrong to add the MiMa exclusion here?

julienrf:
> was I wrong to add the MiMa exclusion here?

Indeed, the correct solution is to duplicate the implicit conversion into the 2.11 and 2.12 source trees, as April said in https://github.com/scala/scala-collection-compat/pull/356#issuecomment-689368290.

NthPortal:
What is the correct thing to do now?

- move the method to the package objects, draft a new release ASAP, and make a note that the release was bad and shouldn't be used?
- leave it as is?
- have the method in both places?

SethTisue:
> What is the correct thing to do now?

Let's discuss on #367

PR #365, opened 2020-09-09
use latest MiMa version in binary-compat test

SethTisue:
this will take some work, the APIs the test was using don't exist anymore

we may need to assess what this test is for and whether we really need it

fyi @dwijnand


SethTisue:
This turned out to be easy, actually. I'm merging it since the changes are superficial and I want to get a release out and this is just test code, but I'll of course respond to any feedback regardless.

PR #366, opened 2020-09-10
clean up build (including a fix to make LazyListGCTest run)

SethTisue:
thank you new sbt 1.4 unused-settings warnings for catching the `LazyListGCTest` thing!!

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -63,9 +63,9 @@ lazy val compat = MultiScalaCrossProject(JSPlatform, JVMPlatform, NativePlatform
         if (scalaVersion.value.startsWith("2.13.")) sharedSourceDir / "scala-2.13"
         else sharedSourceDir / "scala-2.11_2.12"
       },
-      Test / sourceDirectories += (ThisBuild / baseDirectory).value / "compat/src/test/scala-jvm"
     )
     .jvmSettings(
+      Test / unmanagedSourceDirectories += (ThisBuild / baseDirectory).value / "compat/src/test/scala-jvm",
       junit,
       scalaModuleMimaPreviousVersion := Some("2.1.6"),
       mimaBinaryIssueFilters ++= {
--- a/build.sbt
+++ b/build.sbt
@@ -19,8 +19,10 @@ lazy val commonSettings = Seq(
 lazy val root = project
   .in(file("."))
   .settings(commonSettings)
-  .settings(name := "scala-collection-compat")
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -66,26 +60,34 @@ lazy val compat = MultiScalaCrossProject(JSPlatform, JVMPlatform, NativePlatform
...
+      Compile / unmanagedSourceDirectories += {
+        val sharedSourceDir = (ThisBuild / baseDirectory).value / "compat/src/main"
         if (scalaVersion.value.startsWith("2.13.")) sharedSourceDir / "scala-2.13"
         else sharedSourceDir / "scala-2.11_2.12"
       },
-      Test / sourceDirectories += (ThisBuild / baseDirectory).value / "compat/src/test/scala-jvm"
     )
     .jvmSettings(
-      junit
+      Test / unmanagedSourceDirectories += (ThisBuild / baseDirectory).value / "compat/src/test/scala-jvm",
```
NthPortal, line 70: it's not clear to me why `sourceDirectories` doesn't work/what it does, but if this fixes it, great

build.sbt
```diff
@@ -264,14 +266,6 @@ lazy val `scalafix-tests` = project
   .dependsOn(`scalafix-input`, `scalafix-rules`)
   .enablePlugins(BuildInfoPlugin, ScalafixTestkitPlugin)
 
-lazy val dontPublish = Seq(
-  publishArtifact := false,
-  packagedArtifacts := Map.empty,
-  publish := {},
-  publishLocal := {},
-  scalaModuleMimaPreviousVersion := None
```
NthPortal, line 272: is there going to be an issue with this setting being missing where `don'tPublish` is now replaced by `publish / skip := true`?

NthPortal (approved):
LGTM, though I am not by any means an sbt expert

build.sbt
SethTisue: same here :-)

build.sbt
SethTisue: before, `scalaModuleMimaPreviousVersion` was set in `commonSettings` build-wide so then we had to replace that with `None`. why I did it that way originally I don't know — just not really thinking, I think?

but in any case, in the new version we only set it in the specific place where we actually want it (namely in `compat`'s `jvmSettings`), and just let it default to `None` everywhere else

Issue #367, opened 2020-09-15
did 2.2.0 break bincompat on 2.11?


SethTisue:
old tickets that explain the problem are: #288, #292, #249 

PR with the breakage: #341, and then I wrongly added a MiMa exclusion in #356

most recent discussion: #356 (but let's continue here)

on #356 @NthPortal wrote:

>What is the correct thing to do now?
>
> - move the method to the package objects, draft a new release ASAP, and make a note that the release was bad and shouldn't be used?
> - leave it as is?
> - have the method in both places?

SethTisue:
> move the method to the package objects, draft a new release ASAP, and make a note that the release was bad and shouldn't be used

I think this is what we should do. Besides the handful of people active on the tickets, probably only Scala Steward users even noticed the release, and if we do a prompt followup release, the steward will send followup PRs.

But I will bring it up at our team meeting in a few hours and see what others think.

SethTisue:
I just found https://github.com/scala/scala-collection-compat/issues/162, in which @szeiger and @dwijnand both made the exact same argument I made:

> the trait is package-private (which MiMa doesn't see) and should not be extended by users

as a result, releases since 2.1.1 have not been binary compatible with 2.0.0, if we are right that this form of change is not safe

SethTisue:
@julienf are you absolutely, positively certain that the change is not safe, even though the trait is package-private?

lrytz:
I also think the new MiMa rule
```
exclude[ReversedMissingMethodProblem]("scala.collection.compat.PackageShared.*"), // it's package-private
```
is correct. The backwards bin compat issue is that subclasses of the trait need to be recompiled to gain a mixin forwarder method which implements the new abstract method in the interface. All subclasses are known, part of the library itself, so we're fine.

julienrf:
Sorry, I overlooked at the problem, it seems that we are safe!

NthPortal:
does that mean we can deduplicate the methods back into `PackageShared`?

SethTisue:
@julienrf better a surplus of caution than a shortage!

> does that mean we can deduplicate the other methods back into PackageShared?

let's try it: #368

PR #368, opened 2020-09-15
use PackageShared to de-duplicate code

SethTisue:
as per #367, we thought the duplication was necessary for binary
compatibility, but it isn't


SethTisue:
draft because I only did it in one place. if CI likes it we could look for other candidates for de-duplication

SethTisue:
CI likes it. someone else want to take this and run with it?

PR #375, opened 2020-10-14
get admin/scalafmt.sh working again

SethTisue:
get admin/scalafmt.sh working again
    
recently this started failing on Travis-CI


SethTisue:
note that the scalafmt version we're using is antiquated and doesn't like trailing commas in build.sbt.  I didn't attempt an upgrade

the coursier launch script we're using is also considered antiquated, but I didn't attempt to replace it with some other mechanism

perhaps a volunteer would like to try to upgrade either/both

PR #386, opened 2020-11-17
Add cross build for Scala 3

### LINKED ISSUE #385: add a Scala 3 crossbuild
probably it works anyway with `withDottyCompat`, but:

* adding `withDottyCompat` is a bit of extra hassle
* even if `withDottyCompat` _appears_ to work, we don't want users to be wondering whether it _really_ works or just, like, usually works
* it would be good to get an early start on making the code compilable on Scala 3, and perhaps the Scala 3 compiler will find issues the Scala 2 compiler missed
* it would permit adding the repo to the Dotty community build, to increase the chances of catching compiler regressions

and so on.

volunteer?

note that there is almost no code at all in `compat/src/main/scala-2.13`, nearly all of the code in this repo is for 2.11/2.12 only, so the likelihood of running into any difficulty here is pretty small

once done, I'd be happy to take care of adding it to the Dotty community build.

griggt:
A first go at cross building for Scala 3.0.0-M1. Let's see what I missed or broke.

A few notes:

 * The default Scala.js version in `plugins.sbt` was changed to 1.3.1 since this seems to be the only version that works here with all Scala versions 2.11 through 3.0.0-M1. In particular, a fix to support Scala 3 in ScalaJSJUnitPlugin was just recently merged for the 1.3.1 release (scala-js/scala-js#4285)

   I don't know what (if anything) this change will break.

 * Scala 3 disallows ClassTag[Nothing] (lampepfl/dotty#1730)

   The change to ArraySeqTest.check() is the only solution I could come up with in a reasonable amount of time that worked for all supported Scala versions.

   `ArraySeq.empty[T]` would work for Scala 2.13 and Scala 3, but not for Scala 2.11/2.12 since the signature in this library has an `AnyRef` upper bound unlike the 2.13 stdlib.

    Perhaps someone has a better idea of what to do here.

* Scala.js complained about the `@Test(timeout = 10000)` annotation in `LazyListTest` when building with Scala 3:

  ```scala
  [error] 40 |  @Test(timeout = 10000) // scala/bug#6881
  [error]    |        ^^^^^^^^^^^^^^^
  [error]    |10000L is an unsupported argument for the JUnit @Test annotation in this position
  ```

  The error message is confusing because the `L` suffix was omitted but says it's unsupported. I added the `L` suffix, and the error went away.

Fixes #385


project/MultiScalaProject.scala
```diff
@@ -44,6 +44,7 @@ trait MultiScala {
         val sourceDir = (ThisBuild / baseDirectory).value / base / "src" / "main"
         CrossVersion.partialVersion(scalaVersion.value) match {
           case Some((2, n)) if n >= 12 => List(sourceDir / "scala-2.12+")
+          case Some((3, _))            => List(sourceDir / "scala-2.13")
           case _                       => Nil
         }
       },
```
griggt, line 50: I don't believe this change is necessary.

In fact, it's not clear to me that the rest of this block is required as I don't see a `scala-2.12+` directory anywhere.

SethTisue: There is some extra complexity in the build because the repo includes Scalafix rules. I don't really understand that complexity, and perhaps some of it is unnecessary, but I don't think either you nor I needs to worry about it right now, if tests are passing and the community build likes it then :+1:, and the next time we publish if it goes well then :+1:

SethTisue:
Your changes LGTM. Thanks for tackling this!

The CI failures have to do with the 2.13.4 upgrade. I'll take care of that next week, and I'll also publish a new version.

PR #389, opened 2020-11-21
upgrade to Scala 2.13.4

SethTisue:
not for merge until a Scala 3 release is available that upgrades the standard library to 2.13.4 -- presumably that will be 3.0.0-M2. (M1 is on 2.13.3.)

I copied `LazyList.scala` and `LazyListLazinessTest.scala` straight from 2.13.4, except for the package declaration /cc @NthPortal


NthPortal:
> I copied `LazyList.scala` and `LazyListLazinessTest.scala` straight from 2.13.4, except for the package declaration

sadly, the names and APIs are sufficiently different between 2.11/2.12 and 2.13 that I don't think we can do that. also, some of the tests for `LazyList` were commented out basically as a note that 2.11 or 2.12 can't support them, though I have no recollection of why.

I think we're better off trying to cherry-pick the change commits, though I don't really know how given the completely different directory structure (and the different package). I can try to have a look at doing this if I have the energy

SethTisue:
yeah honestly I don't think I can afford to take this on myself, then. hopefully you or another volunteer will choose to tackle it

SethTisue:
the Dotty PR with the 2.13.4 upgrade is https://github.com/lampepfl/dotty/pull/10392 — once it lands, a nightly could be used for testing, even before the next official release

PR #390, opened 2020-11-21
Add a few tools to gitignore

ekrich:


SethTisue:
we generally prefer people put these things in their global ignores, rather than have to make these same changes in literally every Scala repo

OTOH I don't care that much one way or the other ¯\_(ツ)_/¯

ekrich:
I guess people are pretty liberal about this in Scala Native and Scala.js. I put little categories and when it is just a blob then who knows what the ignores are for. I do think it puts a little less burden on the casual contributor but ... up to you.

PR #398, opened 2020-12-05
avoid using type parameters on Left or Right

SethTisue:
this future-proofs the code for scala/scala#9303, and IMO the new code is just as good anyway

the code could be even better with `withRight` (we wouldn't need to repeat `(String, Int)`), but we don't have `withRight` in 2.11/2.12

as for the reformatting, well, scalafmt insisted


NthPortal (approved):
motivation for backporting `withRight`/`withLeft` 👀

SethTisue:
> motivation for backporting withRight/withLeft 👀

that'd be awesome

PR #402, opened 2020-12-19
Cross publish to scala 3.0.0-M3

cquiroz:
I think this PR should allow to publish `scala-collection-compat` for both M2 and M3

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ scala:
   - 2.12.12
   - 2.13.3
   - 3.0.0-M2
+  - 3.0.0-M3
 
 env:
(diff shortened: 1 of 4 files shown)


ekrich:
Gosh, wouldn't it be easier to just bump the Scala version to M3 and then get a new version published.

cquiroz:
Possibly. I'm guessing having M2/M3 maybe useful. At any rate it is failing in travis

ekrich:
This library is very fast at publishing new versions. Surprisingly, I am seeing a `RC1` - https://repo1.maven.org/maven2/org/scala-lang/

SethTisue:
let's go with #403

PR #405, opened 2020-12-20
Scala 2.13.4 (was 2.13.3)

SethTisue:


SethTisue:
@NthPortal CI shows a `LazyList` related test failure, but on Scala.js only — can interest you into looking into what that's about...?

NthPortal:
We need to wait for a version of Scala.js build with 2.13.4 is published before that test will work for Scala.js

SethTisue:
@sjrd are you planning on rolling a new Scala.js version?

sjrd:
The next version will probably be around mid January, about 2 months after the latest release. Definitely not before the week is January 4, as I'm on vacation until then.

SethTisue:
SGTM

PR #417, opened 2021-01-26
make `@nowarn` a ClassfileAnnotation, as in 2.12.13

### LINKED ISSUE #397: Drop 2.12's scala.annotation.nowarn once 2.12.13 is out
See https://github.com/scala/scala-lang/pull/1184#discussion_r534666118

2.12.13 is coming out imminently  (today?).  Once it's out, probably best when bumping to 2.12.13, it would be good to move that noop to 2.11 only.

But two scenarios to think about:

1. users that upgrade `scalaVersion` but not `scala-collection-compat` (or if we don't implement this change): it's unclear what classfile for `@nowarn` they'll class load, which could break the configurable warning feature

2. users that upgrade `scala-collection-compat` (with this implemented) but not `scalaVersion`: won't have a `@nowarn` in they're classpath any more... 😕 🤔 ...

SethTisue:
fixes #397

### DIFF
--- a/compat/src/main/scala-2.11_2.12/scala/annotation/nowarn.scala
+++ b/compat/src/main/scala-2.11_2.12/scala/annotation/nowarn.scala
@@ -12,11 +12,11 @@
 
 package scala.annotation
 
-/** An annotation for local warning suppression added in 2.13.2. Note that this annotation has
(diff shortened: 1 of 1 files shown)


SethTisue:
I did some testing locally via `publishLocal` and then consuming that from a little sample sbt build, and everything seems fine

PR #421, opened 2021-01-30
Deprecate update method on i.ArraySeq impls

### LINKED ISSUE #419: immutable.ArraySeq should not have "update" method
not immutable 😇 

- https://github.com/scala/scala/blob/v2.13.4/src/library/scala/collection/immutable/ArraySeq.scala#L475-L513
- https://github.com/scala/scala-collection-compat/blob/v2.4.0/compat/src/main/scala-2.11_2.12/scala/collection/compat/immutable/ArraySeq.scala#L177

```
Welcome to Scala 2.13.4 (OpenJDK 64-Bit Server VM, Java 1.8.0_275).

scala> val a = new scala.collection.immutable.ArraySeq.ofInt(Array(1, 2, 3))
val a: collection.immutable.ArraySeq.ofInt = ArraySeq(1, 2, 3)

scala> a(0) = 10
       ^
       error: value update is not a member of collection.immutable.ArraySeq.ofInt
       did you mean updated?
```

```
Welcome to Scala 2.12.12 (OpenJDK 64-Bit Server VM, Java 1.8.0_275).

scala> import scala.collection.compat.immutable.ArraySeq
import scala.collection.compat.immutable.ArraySeq

scala> val a = new ArraySeq.ofInt(Array(1, 2, 3))
a: scala.collection.compat.immutable.ArraySeq.ofInt = ArraySeq(1, 2, 3)

scala> a(0) = 10

scala> a
res1: scala.collection.compat.immutable.ArraySeq.ofInt = ArraySeq(10, 2, 3)
```

NthPortal:
Deprecate `update` method on `i.ArraySeq` implementations,
and change the implementations of the method to throw
an `UnsupportedOperationException`.

Fixes #419

### DIFF
--- a/compat/src/main/scala-2.11_2.12/scala/collection/compat/immutable/ArraySeq.scala
+++ b/compat/src/main/scala-2.11_2.12/scala/collection/compat/immutable/ArraySeq.scala
@@ -115,13 +115,21 @@ object ArraySeq {
         ArrayBuilder.make[T]()(m) mapResult ArraySeq.unsafeWrapArray[T]
     }
 
+  private def unsupportedUpdate: Nothing =
+    throw new UnsupportedOperationException("immutable ArraySeq does not support mutation")
+
+  private final val updateDeprecationMsg =
+    "Mutation is not supported for immutable ArraySeq, and should not have ever been " +
+      "part of the API"
+
   @SerialVersionUID(3L)
   final class ofRef[T <: AnyRef](val unsafeArray: Array[T]) extends ArraySeq[T] with Serializable {
     lazy val elemTag         = ClassTag[T](unsafeArray.getClass.getComponentType)
     def length: Int          = unsafeArray.length
     def apply(index: Int): T = unsafeArray(index)
-    def update(index: Int, elem: T) { unsafeArray(index) = elem }
-    override def hashCode = MurmurHash3.arrayHash(unsafeArray, MurmurHash3.seqSeed)
+    @deprecated(updateDeprecationMsg, since = "2.4.2")
+    def update(index: Int, elem: T): Unit = unsupportedUpdate
+    override def hashCode                 = MurmurHash3.arrayHash(unsafeArray, MurmurHash3.seqSeed)
     override def equals(that: Any) = that match {
       case that: ofRef[_] =>
         arrayEquals(unsafeArray.asInstanceOf[Array[AnyRef]],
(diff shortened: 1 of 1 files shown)


compat/src/main/scala-2.11_2.12/scala/collection/compat/immutable/ArraySeq.scala
```diff
@@ -115,13 +115,21 @@ object ArraySeq {
...
+
   @SerialVersionUID(3L)
   final class ofRef[T <: AnyRef](val unsafeArray: Array[T]) extends ArraySeq[T] with Serializable {
     lazy val elemTag         = ClassTag[T](unsafeArray.getClass.getComponentType)
     def length: Int          = unsafeArray.length
     def apply(index: Int): T = unsafeArray(index)
-    def update(index: Int, elem: T) { unsafeArray(index) = elem }
-    override def hashCode = MurmurHash3.arrayHash(unsafeArray, MurmurHash3.seqSeed)
+    @deprecated(updateDeprecationMsg, since = "2.3.3")
+    def update(index: Int, elem: T): Unit = unsupportedUpdate
```
julienrf, line 131: I guess we can assume that nobody has relied on this previous implementation since code calling `update` would not have compiled with Scala 2.13, and the purpose of scala-collection-compat is to cross-compile with 2.12 and 2.13…

So, I think it’s OK to throw an exception here.

compat/src/main/scala-2.11_2.12/scala/collection/compat/immutable/ArraySeq.scala
```diff
@@ -115,13 +115,21 @@ object ArraySeq {
...
+      "part of the API"
+
   @SerialVersionUID(3L)
   final class ofRef[T <: AnyRef](val unsafeArray: Array[T]) extends ArraySeq[T] with Serializable {
     lazy val elemTag         = ClassTag[T](unsafeArray.getClass.getComponentType)
     def length: Int          = unsafeArray.length
     def apply(index: Int): T = unsafeArray(index)
-    def update(index: Int, elem: T) { unsafeArray(index) = elem }
-    override def hashCode = MurmurHash3.arrayHash(unsafeArray, MurmurHash3.seqSeed)
+    @deprecated(updateDeprecationMsg, since = "2.3.3")
```
julienrf, line 130:
```suggestion
    @deprecated(updateDeprecationMsg, since = "2.4.2")
```

julienrf (approved):

compat/src/main/scala-2.11_2.12/scala/collection/compat/immutable/ArraySeq.scala
NthPortal: not only that, but since it's only defined on the implementations (`ofRef`, `ofInt`, etc.) and not `ArraySeq` itself, you would have had to manually create an instance of the implementation type, which is both less convenient and unlikely

SethTisue:
@NthPortal squash, then let's merge?

PR #485, opened 2021-10-01
replace Travis-CI with GitHub Actions

SethTisue:
context: https://github.com/scala/scala-dev/issues/785


SethTisue:
this is still very WIPpy at the moment

this module has the weirdest build and CI setup of any of the modules, so it's going to take some extra thought and care

* [ ] adapt/replace `TEST_SCALAFIX`
* [ ] adapt/replace `TEST_BINARY_COMPAT`
* [ ] adapt/replace `TEST_SCALAFMT` and the `scalafmt-test` and `scalafmt` build commands
  * `admin` directory had `install-scalafmt-githook.sh` and `pre-commit` and `scalafmt.sh` but I really don't want all that, surely we can do it some simpler way?
* [ ] replace `ci` build command
* [ ] I downgraded to scalafix 0.9.30 because with 0.9.31 it hit some 2.12.14 vs 2.12.15 I didn't understand or dig into yet

SethTisue:
frustrated. abandoning this approach of trying to restructure things. I'll make a new PR that just keeps everything the same and changes CI systems only

SethTisue:
resubmitting some of this stuff at #486, while still staying on Travis-CI

then we can switch CI systems in another followup PR

SethTisue:
note that the problems I thought I was having with the scalafix version were actually because because setting `excludeFilter` to make `headerCheck` pass was breaking `scalafixTests`. in #487 I'm skipping `headerCheck` when it's unwanted with `headerCheck := Nil` instead

PR #495, opened 2021-10-29
preserve immutable collections in to(Target) conversions

lrytz:


compat/src/main/scala-2.11_2.12/scala/collection/compat/CompatImpl.scala
```diff
@@ -14,36 +14,35 @@ package scala.collection.compat
...
 
   private[this] def ruin(): Unit = {
     if (collection != null) that ++= collection
-    collection = null.asInstanceOf[CC[A]]
+    collection = null.asInstanceOf[C]
     ruined = true
   }
 
   override def ++=(elems: TraversableOnce[A]): this.type =
     elems match {
```
johnynek, line 38: nit. I assume the `ct` unapply happens before the predicate check, but those are very fast and this branch can only be taken once. What about:
```scala
@inline def proxy(): Unit = {
  ruin()
  that ++= elems
}

if ((collection eq null) && !ruined) {
  elems match {
    case ct(ca) =>
      collection = ca 
    case _ => proxy()
}
else proxy()

this
```

compat/src/main/scala-2.11_2.12/scala/collection/compat/CompatImpl.scala
```diff
@@ -53,14 +52,29 @@ private final class IdentityPreservingBuilder[A, CC[X] <: TraversableOnce[X]](
   }
 
   def clear(): Unit = {
-    collection = null.asInstanceOf[CC[A]]
+    collection = null.asInstanceOf[C]
     if (ruined) that.clear()
```
johnynek, line 56:
```scala
if (ruined) {
  that.clear()
  ruined = false
}
```
since the else branch already has `ruined == false`

compat/src/main/scala-2.11_2.12/scala/collection/compat/CompatImpl.scala
lrytz: 👍  thanks

compat/src/main/scala-2.11_2.12/scala/collection/compat/CompatImpl.scala
```diff
@@ -53,14 +51,31 @@ private final class IdentityPreservingBuilder[A, CC[X] <: TraversableOnce[X]](
   }
 
   def clear(): Unit = {
-    collection = null.asInstanceOf[CC[A]]
-    if (ruined) that.clear()
-    ruined = false
+    collection = null.asInstanceOf[C]
+    if (ruined) {
+      that.clear()
```
NthPortal, line 56: is there also an invariant "`!ruined` => `that` is empty"?

lrytz: yes, elements are only added to `that` after `ruin()`

NthPortal (commented):
initial review is that it looks correct. if I get more energy, hopefully I can do a more thorough review.

I did not review of the sbt changes, as I'm less confident in my ability to refactor sbt configs than in yours

lrytz:
Thank you @NthPortal!

johnynek (approved):

retronym (approved):

SethTisue:
@lrytz can you fill in the PR description?

lrytz:
👍  done

NthPortal:
oh, this is already released, awesome!

PR #498, opened 2021-11-09
set version policy for 2.5.1 release

SethTisue:


SethTisue:
bah. I'll just go to version 2.6.0

Issue #355, opened 2020-08-13
Mutable updateWith missing


nilskp:
`scala.collection.mutable.MapOps.updateWith`

SethTisue:
fixed by #508

PR #508, opened 2021-12-21
Add mutable Map.updateWith

emanb29:
References https://github.com/scala/scala-collection-compat/issues/355


julienrf (approved):
Thank you!

PR #524, opened 2022-03-07
Add a Scala 3 native build

JD557:
Adds Scala 3 support to the Scala native build.

Scala native only works on 3.1.x+, so this adds a 3.1 project just for native (I also needed to add the `JVMPlatform`, since otherwise the crossproject wouldn't work, but it's not referenced later).

This breaks the MIMA tests (as there's no previous version). I added a quick fix to skip the tests on this version.

Supercedes #523.


SethTisue:
Thank you!

Let's pursue publishing here: #525

Issue #532, opened 2022-05-18
scala.jdk.OptionConverters missing?


haukeh:
Greetings, 

while cross-building with Scala 2.12 and 2.13., I just stumbled upon the fact, that there is no `scala.jdk.OptionConverters` in this library. As `CollectionConverters` is present, I am wondering if there is a specific reason why `OptionConverters` was ommited.  

Prior to Scala 2.13 the standard way to go (I guess) was using [scala-java8-compat](https://github.com/scala/scala-java8-compat)'s `scala.compat.java8.OptionConverters`, so probably the idea was that most people would just keep using that instead of the 2.13 `scala.jdk` import? 

Sorry if this is a weird question. It would be great to hear your opinions on this. 

Best regards,
Hauke

lrytz:
I think it's an oversight / nobody asked or thought about it. https://github.com/scala/scala-collection-compat/pull/217 for reference. A contribution would be welcome!

haukeh:
Thanks for the quick reply, Lukas. I'd be willing to add the missing class(es) and create a PR, in fact I've already tried yesterday evening, but I'm struggling a little bit with the `MultiScalaCrossProject`. As `OptionConverters` references `java.lang.Optional` it would need to be included only for the JVM and for Scala 2.11/2.12. I'm still trying to figure out, in which source directory it would need to be for that to work.

PR #533, opened 2022-05-19
Backport scala.jdk.OptionConverters to 2.11 / 2.12

haukeh:
This addresses #532  

I have simply copied the implementations of `OptionConverters` and `OptionShape` from the standard library, including license string, scaladoc (minus references made to `javaapi`, which doesn't exist here). A quick look at the git history indicated that this has been done before. I'm not sure if this raises any issues with the license or anything else I'm not aware of, so please let me know if there would be a better way to go about this!

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -95,6 +95,15 @@ lazy val compat = new MultiScalaCrossProject(
     )
     .jvmSettings(
       Test / unmanagedSourceDirectories += (ThisBuild / baseDirectory).value / "compat/src/test/scala-jvm",
+      Compile / unmanagedSourceDirectories += {
+        val jvmParent = (ThisBuild / baseDirectory).value / "compat/jvm/src/main"
+        CrossVersion.partialVersion(scalaVersion.value) match {
+          case Some((3, _) | (2, 13)) =>
+            jvmParent / "scala-2.13"
+          case _ =>
+            jvmParent / "scala-2.11_2.12"
+        }
+      },
       junit,
     )
     .disablePlugins(ScalafixPlugin),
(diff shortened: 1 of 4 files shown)


lrytz:
Hi @haukeh, I'm sorry for the delay.

I think `java.lang.Optional` is available on bot Scala.js and Scala Native. As far as I know they both use the scala re-implementation of the JDK, which contains Optional (https://github.com/scala-js/scala-js/blob/main/javalib/src/main/scala/java/util/Optional.scala).

haukeh:
Hi @lrytz, thanks for the info! When I first tried adding it to the default source root, some JS tests failed with errors like:

```
[error] Referring to non-existent method static java.util.OptionalDouble.of(double)java.util.OptionalDouble
```

Which made me - having literally no idea about scala-js - just assume `Optional` in general is not available. But now, after taking a closer look, it seems that it's only the primitive Optional wrappers like `OptionalDouble` etc. that are not available in JS. 

So I guess a way to solve this would be creating a "lightweight" version of `OptionConverters` in the js and native source folders, which only contains the `Option[T]` stuff?

haukeh:
Ok, so I just found out that `java.util.Optional` is available in Scala-Native on the main branch, but it has not been released yet. So for now I opted to create a slim variant of OptionConverters for JS only. Let me know what you think.

lrytz:
> I just found out that `java.util.Optional` is available in Scala-Native on the main branch

cc @WojciechMazur -- is that possibly changing soon?

Otherwise I'm happy to go with the current solution. Thank you @haukeh!

WojciechMazur:
@lrytz There should be a new release of Scala Native introducing j.u.Optional later this week or at the beginning of the next one. I'm currently on mini-vacations and cut off from my workstation. 
The implementation of Optional is exactly the same as on the Scala.js so it should work just fine

PR #544, opened 2022-07-08
bump Scala 2.12 version to 2.12.16 (was .15)

SethTisue:


SethTisue:
oops, #535 is blocking this

SethTisue:
oh, ha, actually #535 isn't blocking this!

the problem went away when I changed it so we do e.g. `++2.12.x` instead of `++2.12.x!`. The exclamation point in the latter was overriding the value of `scalafixScala212`. Pretty sure that must have been accidental.

PR #547, opened 2022-07-08
for Scala 3 users, require Scala 3.1+

SethTisue:
as per scala/scala-lang#1387, there's no longer any need to keep supporting Scala 3.0.x. it's fine to expect people to upgrade to 3.1 or higher

Admittedly I'm making a bit of an executive decision here. But I see this happening in many places in the ecosystem.

Also, it simplifies the build. Scala 3.0.x didn't support Scala Native, so we had to do weird stuff to publish for 3.0 on JS and JVM but 3.1 on Native. This removes those complications.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,45 +11,43 @@ jobs:
       fail-fast: false
       matrix:
         java: [8, 11, 17]
-        scala: [2.11.x, 2.12.x, 2.13.x, 3.0.x]
-        platform: [jvm, js, native]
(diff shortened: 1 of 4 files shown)


SethTisue:
Well, I'm making a bit of an executive decision here. But I see this happening in many places in the ecosystem.

Also, it simplifies the build. Scala 3.0.x didn't support Scala Native, so we had to do weird stuff to publish for 3.0 on JS and JVM but 3.1 on Native. This removes those complications.

PR #549, opened 2022-07-15
scalafix 0.10.1

bjaglin:
https://github.com/scalameta/scalameta/pull/2684, first released in scalameta 4.5.1 (scalafix 0.10.0), changed a behavior, so this takes a more intrusive but also more robust approach so that the rule works with scalafix 0.10.x.


SethTisue:
thank you!

PR #581, opened 2023-02-13
Make ArraySeq extends IndexedSeqOptimized

RustedBones:
Make `ArraySeq[T]` extends `IndexedSeqOptimized[T, ArraySeq[T]]` so that operation expecting `CanBuildFrom[ArraySeq[_], T, ArraySeq[T]]` return specialized `ArraySeq[T]` instead of `IndexedSeq[T]`.

Without this, the following can't compile
```scala
val x = ArraySeq(1, 2)
val y = ArraySeq(3, 4)
val z: ArraySeq[Int] = x ++ y
```


SethTisue:
@scala/collections ?

RustedBones:
Any news on this one ?

SethTisue:
@lrytz @julienrf maybe one of you would like to review this...?

julienrf (approved):
That looks good to me, and it should also bring performance improvements!

PR #584, opened 2023-03-13
Upgrade Scala Native to 0.4.10 and drop Scala 2.11 support

ekrich:
As mentioned in https://github.com/scala/scala-collection-compat/pull/579, a minor version bump on release would be great.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,6 +15,8 @@ jobs:
         platform: [JVM, JS, Native]
         mode: [normal]
         exclude:
+          - scala: 2.11.x
+            platform: Native
           - java: 11
             platform: JS
           - java: 11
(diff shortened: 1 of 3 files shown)


SethTisue:
I think in order to produce a change of heart, you need to make a case that there is a concrete benefit for Scala Native users and we're not just mechanically taking the upgrade.

(As before, I lean towards your position, but I'm not the decider.)

ekrich:
Ok, no worries. I believe we should drop `2.11` for JVM for newer versions of this library. If people are upgrading and need this library for compatibility to afford the upgrade then they can use any version prior to the drop. Now that Scala Native and Scala.js have dropped support I think this is a good case. Besides, upgrading to `2.12` from normal non cross projects is just a stepping stone to `2.13` and eventually `3.0`. Maintaining backward compat is nice to have for people upgrading Scala.js but doesn't help as much for Scala Native and then we have no regression tests with this library as part of the build and verification.

PR #587, opened 2023-03-31
Split up Using.scala into 2.11 and 2.12 versions

kwalcock:
This is so that a Source can be automatically released in 2.11


compat/src/main/scala-2.11/scala/util/Using.scala
```diff
@@ -0,0 +1,409 @@
...
+    implicit object AutoCloseableIsReleasable extends Releasable[AutoCloseable] {
+      def release(resource: AutoCloseable): Unit = resource.close()
+    }
+
+    /** An implicit `Releasable` for [[scala.io.Source `Source`s]] which aren't [[java.lang.AutoCloseable `AutoCloseable`s]] in Scala 2.11. */
+    implicit object SourceReleasable extends Releasable[Source] {
+      def release(resource: Source): Unit = resource.close()
+    }
+  }
+
```
preveen-stack, line 408: guess we can remove this empty line

kwalcock: It was there before, so I left it in.  Can it be configured in scalafmt if it is a problem?

SethTisue: it's fine to clean up formatting in the course of making changes, but it's also fine not to

SethTisue:
Thank you!

Issue #543, opened 2022-07-08
add `distinctBy`


xuwei-k:

xuwei-k:
🤔 

```diff
--- a/compat/src/main/scala-2.11_2.12/scala/collection/compat/PackageShared.scala
+++ b/compat/src/main/scala-2.11_2.12/scala/collection/compat/PackageShared.scala
@@ -467,6 +467,20 @@ private object SizeCompareImpl {
 
 class TraversableLikeExtensionMethods[A, Repr](private val self: c.GenTraversableLike[A, Repr])
     extends AnyVal {
+
+  def distinctBy[B, That](f: A => B)(implicit cbf: CanBuildFrom[Repr, A, That]): That = {
+    val builder = cbf()
+    val keys = collection.mutable.Set.empty[B]
+    for (element <- self) {
+      val key = f(element)
+      if (!keys(key)) {
+        builder += element
+        keys += key
+      }
+    }
+    builder.result()
+  }
```

stewSquared:
> 🤔

Is this is all there is to it (plus tests)?
I'm happy to make the PR as part of the current spree; it just seems odd that it wouldn't be done yet.

alexklibisz:
First pass here: https://github.com/scala/scala-collection-compat/pull/602

PR #602, opened 2023-06-10
Add implementation of distinctBy

alexklibisz:
Addressing #543 

Shamelessly lifted @xuwei-k 's recommended implementation and added a few unit tests. Happy to expand as necessary.


julienrf:
It is not clear to me why the binary compatibility tests are failing. It seems the value of `versionPolicyIntention` is `Compatibility.BinaryAndSourceCompatible` but I can’t see where we provide such value in the build. @SethTisue do you have an idea?

SethTisue:
the MiMa thing went away on rebase

I also did `scalafmt compat/src/main/scala-2.11_2.12/scala/collection/compat/PackageShared.scala` to fix the scalafmt check

SethTisue (approved):

SethTisue:
thank you @alexklibisz ! I'll roll a release before too much longer (ticket on that is #603)

PR #604, opened 2023-06-14
upgrade Scalafmt (from 1.5.1 to 3.7.4) and reformat sources

SethTisue:
I spent a certain amount of time tinkering with `.scalafmt.conf` trying to make the diff smaller

@julienrf do you see any opportunities to minimize the diff further...?


SethTisue:
I've pushed another commit that switches `align.preset` from `more` to `none`

`more` does a lot more than it did in Scalafmt 1.x, and it seems overbearing to me and likely to result in annoying diffs in the future. so while we're reformatting anyway, let's disable it and pay the costs in present diffs

SethTisue:
there's a few places where it's inserting an annoying newline after `implicit`, and I fooled with `newlines.implicitParamListModifierXXX` and wasn't able to improve it, so I think we can let that go

there were certainly some cases where the alignment was nice-looking, but a lot of others that I found annoying... and most tooling is not likely to preserve the alignment anyway, during development

many of the remaining diffs insert a space to change `//foo` to `// foo`, which I can live with

marking as ready for review

Issue #610, opened 2023-07-06
`unfold` method for `Iterable` companion sub classes


xuwei-k:
https://github.com/scala/scala/blob/57b8dcc27e4769c482a82d4eac4f2781c7d43515/src/library/scala/collection/Factory.scala#L124

PR #611, opened 2023-07-06
add `unfold` method

xuwei-k:
- fix https://github.com/scala/scala-collection-compat/issues/610


SethTisue:
rather than adding MiMa exclusions, `versionPolicyIntention` should be changed from `BinaryAndSourceCompatible` to `BinaryCompatible` (and we'll bump the minor version on next release)

SethTisue:
review by @scala/collections ?

SethTisue (approved):

PR #646, opened 2024-03-18
Upgrade to 0.5.0 and drop Scala 2.11 for Scala Native

ekrich:
There is a level of consensus in the Scala Native community for dropping `0.4.x` when `0.5.0` final is released. This puts developers of libraries in a bit of an awkward situation which we understand. For perspective, we supported `0.3.x` along with `0.4.0-M2` for a long time and it was a bit of a pain for builds so this may be part of the current thinking. Also, this time around we have an active project whereas previously it was unknown when `0.4.0` final would occur.

- Adds Scala Native `0.5.0` and drops support for Scala 2.11.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,8 @@ jobs:
             platform: JS
           - java: 17
             platform: Native
+          - scala: 2.11.x
+            platform: Native
         include:
           - java: 8
             scala: 2.12.x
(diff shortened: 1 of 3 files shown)


lolgab:
I think you can avoid adding Scala 2.11, nobody is using it.

ekrich:
I removed it from CI and the build.

PR #658, opened 2024-08-09
Add Array.from extension for scala 2.11 and 2.12

RustedBones:
Without the extension, code fails to compile for scala 2.11 and 2.12 with
```
 value from is not a member of object Array
     val a = Array.from(xs)
``` 

Add missing extension to the compat package

### DIFF
--- a/compat/src/main/scala-2.11_2.12/scala/collection/compat/PackageShared.scala
+++ b/compat/src/main/scala-2.11_2.12/scala/collection/compat/PackageShared.scala
@@ -362,7 +362,10 @@ private[compat] trait PackageShared {
 
 class ArrayExtensions(private val fact: Array.type) extends AnyVal {
   def from[A: ClassTag](source: TraversableOnce[A]): Array[A] =
-    fact.apply(source.toSeq: _*)
+    source match {
+      case it: Iterable[A] => it.toArray[A]
+      case _ => source.toIterator.toArray[A]
+    }
 }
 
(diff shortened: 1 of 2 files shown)


SethTisue:
review by @scala/collections ?

compat/src/main/scala-2.11_2.12/scala/collection/compat/PackageShared.scala
```diff
@@ -357,6 +360,11 @@ private[compat] trait PackageShared {
     new RandomExtensions(self)
 }
 
+class ArrayExtensions(private val fact: Array.type) extends AnyVal {
+  def from[A: ClassTag](source: TraversableOnce[A]): Array[A] =
+    fact.apply(source.toSeq: _*)
```
He-Pin, line 365: source match {
    case it: Iterable[A] => it.toArray[A]
    case _ => it.iterator.toArray[A]
  }

compat/src/main/scala-2.11_2.12/scala/collection/compat/PackageShared.scala
```diff
@@ -357,6 +360,11 @@ private[compat] trait PackageShared {
     new RandomExtensions(self)
 }
 
+class ArrayExtensions(private val fact: Array.type) extends AnyVal {
```
He-Pin, line 363: final class?

SethTisue (approved):

SethTisue:
Thank you!

SethTisue:
this has now shipped: https://github.com/scala/scala-collection-compat/releases/tag/v2.13.0

PR #660, opened 2024-09-03
Scala 3.3.3, 2.13.13, 2.12.19

SethTisue:
I didn't go to 2.13.14 because I'd have to upgrade Scala.js as well and we seek to avoid doing that in this repo unless we must, for maximum compatibility

going from Scala 3.2 to 3.3 might seem inconsistent with that, but I think we should do it, because all of the messaging around Scala 3 has been super consistent that 3.3 is the first LTS version and older versions are dead. besides, 3.3 isn't at all new anymore.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -50,9 +50,9 @@ lazy val root = project
 lazy val junit = libraryDependencies += "com.github.sbt" % "junit-interface" % "0.13.3" % Test
 
 lazy val scala211 = "2.11.12"
-lazy val scala212 = "2.12.17"
-lazy val scala213 = "2.13.11"
-lazy val scala3 = "3.2.2"
+lazy val scala212 = "2.12.19"
+lazy val scala213 = "2.13.13"
+lazy val scala3 = "3.3.3"
 
 lazy val compat = new MultiScalaCrossProject(
   "compat",


manuzhang:
Thanks @SethTisue. Do we have plan to make a new release after this update?

SethTisue:
Nothing substantive has been merged since 2.12.0, so I don't see a reason to.

SethTisue:
Oh, there's #658. We could release after that one.

manuzhang:
Thanks @SethTisue. Since JDK 21 support is added into 2.12.18, this upgrade (and next release) really helps downstream projects to simplify Scala dependencies.

SethTisue:
hmm, but I don’t think it does; it doesn’t matter in the slightest to downstream users whether we upgraded our Scala version here or not

manuzhang:
Do you mean we shouldn't pull in Scala dependencies from this project?

SethTisue:
@manuzhang I don't understand your question

manuzhang:
Any plan to make a new release soon?

SethTisue:
> Any plan to make a new release soon?

Not sure if it's been considered. Do you see a specific reason to?

SethTisue:
Hmm, I guess we could do one, yeah — but the motivation would be to ship #658, and not anything having to do with this PR.

PR #678, opened 2025-01-20
Update README.md with ScalaJs release badge

joriscode:
I think having ScalaJs badge would highlight that ScalaJs is supported.


SethTisue:
mind signing the Scala CLA?

SethTisue:
Also, I'm wondering if we want to just rid of the JVM badge altogether.

I think these badges (and the related practice of putting the current version number in the README directly) got started in an era when the GitHub web UI wasn't very good at surfacing the library's current version number. These days, the GitHub UI does a better job of that.

So there was a sort of badges era, which I feel has now passed. At least, that's my own subjective impression.

joriscode:
> Note also that we published 2.13.0 an hour or two ago, but the badge still says 2.12.0, I guess because Scaladex takes time to update. (How much time — hours, days? I don't know.)

(One hour later) It is up-to-date now :)

> But I'm wondering if we want to just rid of the JVM badge altogether, rather than add more badges.

On the bright side, it demonstrates the different targets (JVM, JS, Native), but this comes with the downside that a third badge should be added for Scala Native. It is up-to-you.

SethTisue:
Tell you what, I'll submit a different PR that removes the badge, but also adds a bit  to the readme that emphasizes that JVM, Scala.js, and Scala Native are all supported.

SethTisue:
https://github.com/scala/scala-collection-compat/pull/680

joriscode:
That works for me and thank you for your time!

PR #712, opened 2025-09-30
update `Using` from Scala 2.13.17 and add `UsingTest`

### LINKED ISSUE #692: Update `Using` for Scala 2.13.17
both `Using.scala` and `UsingTest.scala` should be updated once Scala 2.13.17 is out, to pull in the changes from https://github.com/scala/scala/pull/11000

this came up in the Scala 2 community build

SethTisue:
ports https://github.com/scala/scala/pull/11000

fixes #692 

sequel to @NthPortal's #319

I was tempted to drop Scala 2.11, but in the end I was able to keep it in the build by special-casing it to use an older Scala.js version

I also had to do a bit of monkey business in the build for the tests to pass on Scala 3 even though Scala 3 hasn't taken the 2.13.17 upgrade yet. Once they take it, the monkey business can be removed and `UsingTest.scala` can be moved out of `src-jvm` and into `src` (since for some reason, the monkey business only worked on the JVM, not JS or Native).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,13 +11,18 @@ jobs:
       fail-fast: false
       matrix:
         java: [8, 25]
-        scala: [2.11.x, 2.12.x, 2.13.x, 3.x]
+        scala: [2.12.x, 2.13.x, 3.x]
+        scala-js: [""]
         platform: [JVM, JS, Native]
         mode: [normal]
-        exclude:
-          - scala: 2.11.x
-            platform: Native
         include:
+          - java: 8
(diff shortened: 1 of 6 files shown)


SethTisue:
note to self: shall I also update the CI matrix to include 25? maybe 8 and 25 would be a sufficient covering set

SethTisue:
I'm going to go ahead and merge and release this on my own authority, since we want to have a new version available at the same time that 2.13.17 is announced.

SethTisue:
followup to fix publishing: #714

